PR #14, opened 2018-10-22
Fix container source.

n3wscott:
There are some issues remaining in the source. Need to get the CI working.

## Proposed Changes

  *
  *
  *

**Release Note**

```release-note
```


n3wscott:
Working now. Kubebuilder does not get you setup out of the box for clientset generation.

/assign @grantr

pkg/apis/sources/v1alpha1/register.go
```diff
@@ -20,11 +20,16 @@ limitations under the License.
 // +k8s:openapi-gen=true
 // +k8s:deepcopy-gen=package,register
 // +k8s:conversion-gen=github.com/knative/eventing-sources/pkg/apis/sources
+// +genclient
```
grantr: Don't the `+genclient` and `+k8s:deepcopy-gen` directives normally go above the type structs? Why move them here?

See https://github.com/kubernetes-sigs/kubebuilder/blob/804f002125094de37d15e758eec54e751843509a/test/project/pkg/apis/ship/v1beta1/register.go#L20-L24 and https://github.com/kubernetes-sigs/kubebuilder/blob/804f002125094de37d15e758eec54e751843509a/test/project/pkg/apis/ship/v1beta1/frigate_types.go#L38-L39

pkg/apis/sources/v1alpha1/register.go
```diff
@@ -35,4 +40,21 @@ var (
...
+	AddToScheme = SchemeBuilder.AddToScheme
 )
+
+// Resource takes an unqualified resource and returns a Group qualified GroupResource
+func Resource(resource string) schema.GroupResource {
+	return SchemeGroupVersion.WithResource(resource).GroupResource()
+}
+
+// Adds the list of known types to Scheme.
+func addKnownTypes(scheme *runtime.Scheme) error {
```
grantr: Why is this function necessary? It's not present in https://github.com/kubernetes-sigs/kubebuilder/blob/804f002125094de37d15e758eec54e751843509a/test/project/pkg/apis/ship/v1beta1/register.go.

Makefile
```diff
@@ -44,6 +44,7 @@ vet:
 # Generate code
 generate:
 	go generate ./pkg/... ./cmd/...
+	/usr/local/bin/bash  ./hack/update-codegen.sh
```
grantr: Can you rely on the shebang in `./hack/update-codegen.sh` instead? Most people won't have `/usr/local/bin/bash`.

pkg/apis/sources/v1alpha1/register.go
n3wscott: ah yes thanks.

pkg/apis/sources/v1alpha1/register.go
n3wscott: This is was the listers look for by default.

n3wscott: wait, my mistake, I don't need this.

Makefile
n3wscott: Thanks for this reminder. I fixed it to work with linux and mac.

n3wscott:
> Strange that register.go doesn't work out of the box!

It only does not because we create the pkg/client folder with the `./hack/update-codegen.sh`

PR #16, opened 2018-10-22
Working with what kubebuilder is doing, document how to deploy to k8s

n3wscott:
**Release Note**
```release-note
NONE
```


config/default.yaml
```diff
@@ -0,0 +1,165 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: knative-sources
```
matzew: :+1: 

I think that `knative-eventing-source` is a mouthful...

n3wscott:
This now contains all the hacks I had to do to get what kubebuilder gave me to work with nice things like `ko`. This is not all the work that needs to happen, but I am at workable demo state.

pkg/apis/sources/v1alpha1/containersource_types.go
```diff
@@ -76,7 +76,7 @@ type ContainerSourceStatus struct {
 	SinkURI string `json:"sinkUri,omitempty"`
 }
 
-// GetCondition returns the condition currently associated with the given type, or nil.
+// GetCondition returns the condition currently associWantPresentated with the given type, or nil.
```
matzew: wat?

samples/sources_v1alpha1_containersources.yaml
```diff
@@ -9,3 +9,8 @@ spec:
   args:
    - '--label="<3"'
    - '--period=2'
+# Fill this out:
```
matzew: :+1: :smile:

pkg/apis/sources/v1alpha1/containersource_types.go
n3wscott: fixed.

Gopkg.toml
```diff
@@ -44,17 +44,18 @@ required = [
   name = "k8s.io/client-go"
   version = "kubernetes-1.11.3"
 
+[[constraint]]
+  name = "sigs.k8s.io/controller-tools"
+  source = "https://github.com/n3wscott/controller-tools.git"
```
vaikas-google: is there a reason why this is forked? If so, is there a path forward where we can use the non-forked version? Perhaps a comment / issue to track this would be in order. Just to make sure that we're not taking on something that will be forked forevs.

n3wscott:
> readme.md is listed as a binary file, is there a reason for that?

I wish I knew why. I can't see why it is. `-rw-r--r--@  1 nicholss  eng   2210 Oct 23 12:23 README.md`

pkg/controller/containersource/reconcile.go
```diff
@@ -50,6 +51,8 @@ type reconciler struct {
 	recorder      record.EventRecorder
 }
 
+const tryTargetable = true
```
vaikas-google: comment here might be helpful. I assume it means that we first try it as sinkable, then targetable?

pkg/controller/containersource/reconcile.go
```diff
@@ -64,13 +66,13 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 	// See if the source has been deleted
 	accessor, err := meta.Accessor(source)
 	if err != nil {
-		logger.Warnf("Failed to get metadata accessor: %s", err)
+		logger.Warnf("Failed to get metadata accessor: %s", zap.Error(err))
 		return object, err
 	}
 	// No need to reconcile if the source has been marked for deletion.
 	deletionTimestamp := accessor.GetDeletionTimestamp()
 	if deletionTimestamp != nil {
```
vaikas-google: Is there no plans to use finalizers? Don't we need to run some code to remove the source at some point?

vaikas-google: As in the old sources used start/stop model, and in those cases you'd run some code to actually undo what the subscription did.

n3wscott: I will test this but I think because I am owning the deployment, the deployment will be deleted if the source is.

Gopkg.toml
n3wscott: added a comment.

pkg/controller/containersource/reconcile.go
n3wscott: added a comment.

pkg/controller/containersource/reconcile.go
n3wscott: Cascade delete works.

grantr: If the deployment creates external resources, how will it know to delete those external resources when the source is deleted? I don't think it can use termination signals, because those might be generated by the node lifecycle.

We could punt on this by saying that the container source won't be able to manage external resources.

pkg/controller/containersource/reconcile.go
```diff
@@ -133,29 +135,46 @@ func (r *reconciler) getSinkUri(ctx context.Context, source *v1alpha1.ContainerS
...
 	err = duck.FromUnstructured(obj, &t)
 	if err != nil {
-		logger.Warnf("Failed to deserialize sink: %s", err)
+		logger.Warnf("Failed to deserialize sink: %s", zap.Error(err))
 		return "", err
 	}
 
 	if t.Status.Sinkable != nil {
-		return t.Status.Sinkable.DomainInternal, nil
+		return fmt.Sprintf("http://%s/", t.Status.Sinkable.DomainInternal), nil
```
grantr: Strange that http is hardcoded here. Is cluster-internal traffic always https capable?

pkg/controller/containersource/reconcile.go
```diff
@@ -133,29 +135,46 @@ func (r *reconciler) getSinkUri(ctx context.Context, source *v1alpha1.ContainerS
...
 
 // fetchObjectReference fetches an object based on ObjectReference.
-func (r *reconciler) fetchObjectReference(namespace string, ref *corev1.ObjectReference) (duck.Marshalable, error) {
+func (r *reconciler) fetchObjectReference(ctx context.Context, namespace string, ref *corev1.ObjectReference) (duck.Marshalable, error) {
+	logger := logging.FromContext(ctx)
+
 	resourceClient, err := r.CreateResourceInterface(namespace, ref)
 	if err != nil {
-		//glog.Warningf("failed to create dynamic client resource: %v", err)
+		logger.Warnf("failed to create dynamic client resource: %v", zap.Error(err))
```
grantr: Could we just return the error here? The caller also logs errors at https://github.com/knative/eventing-sources/pull/16/files#diff-28bc3fe544a2ddf18f611c0b8e4ac2c8R138. That also eliminates the ctx arg.

pkg/controller/containersource/reconcile.go
```diff
@@ -192,7 +210,7 @@ func (r *reconciler) getDeployment(ctx context.Context, source *v1alpha1.Contain
 		},
 		list)
 	if err != nil {
-		logger.Errorf("Unable to list deployments: %v", err)
+		logger.Errorf("Unable to list deployments: %v", zap.Error(err))
```
grantr: Could this just return the error and let the caller log it?

pkg/controller/containersource/reconcile.go
vaikas-google: Right, what @grantr says is what I tried to say. It's not enough to just kill the deployment, there needs to be a way to say, please undo the "subscribe" that you did before. Kind of like what the previous sources did, they kept enough of the context around that was passed back during unsubscribe.

pkg/controller/containersource/reconcile.go
n3wscott: We will have to revisit this when we decide how we will support paths. But at the moment, all domains == http endpoint on port :80 from what I can see. Let me know if this is incorrect.

pkg/controller/containersource/reconcile.go
n3wscott: OOHHHH!! I see. yes. This is a good idea. I made https://github.com/knative/eventing-sources/issues/21 to solve it.

Gopkg.lock
```diff
@@ -609,15 +609,15 @@
...
   name = "k8s.io/apiextensions-apiserver"
   packages = [
     "pkg/apis/apiextensions",
     "pkg/apis/apiextensions/v1beta1",
   ]
   pruneopts = "T"
-  revision = "408db4a50408e2149acbd657bceb2480c13cb0a4"
-  version = "kubernetes-1.11.2"
+  revision = "4347b330d0ff094db860f2f75fa725b4f4b53618"
+  version = "kubernetes-1.10.1"
```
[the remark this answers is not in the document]
n3wscott: I have no idea.... KUBEBUILDER!?!!?!

pkg/controller/containersource/reconcile.go
grantr: Created #22

Gopkg.toml
grantr: Is there an issue tracking this in the kubebuilder repo?

pkg/controller/containersource/reconcile.go
```diff
@@ -106,9 +111,11 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 			r.recorder.Eventf(source, corev1.EventTypeNormal, "Deployed", "Created deployment %q", deploy.Name)
 			source.Status.MarkDeploying("Deploying", "Created deployment %s", args.Name)
```
grantr: Optional suggestion: what do you think about adding a `return source, nil` here, since this is effectively a terminal state for this reconcile pass? That allows the following `else` block to be elevated up one level, which seems closer to idiomatic go style.

PR #24, opened 2018-10-24
Add rbac annotations to ContainerSource controller

grantr:
Currently ContainerSource controller needs to have full permissions on Deployment and ContainerSource resources. It also needs to read (get, list, watch) Sinkable resources. The default added here is Knative Services.

Also ran `hack/update-codegen.sh && hack/update-manifests.sh` to update the generated yaml files.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -95,6 +95,14 @@ rules:
   - patch
   - delete
 - apiGroups:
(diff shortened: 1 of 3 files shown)


n3wscott:
/lgtm

Thanks for catching this.

PR #26, opened 2018-10-24
Kubernetes event source controller

grantr:
Fixes https://github.com/knative/eventing/issues/447

## Proposed Changes

  * Create a `KubernetesEventSource` CRD in `sources.eventing.knative.dev`
  * Add a skeleton controller implementation

This is only the CRD and controller. The receive adapter for https://github.com/knative/eventing/issues/446 will be a separate PR.

**Release Note**

```release-note
New source KubernetesEventSource: Watches Kubernetes corev1.Event resources and sinks them to a Sinkable.
```


pkg/controller/kuberneteseventsource/reconcile.go
```diff
@@ -0,0 +1,197 @@
...
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// TODO(n3wscott): To show the source is working, and while knative eventing is
+// defining the ducktypes, tryTargetable allows us to point to a Service.service
```
vaikas-google: I think this comment is slightly wrong. It's saying that we'll try Targetable. It's not really only Service.service (which I'm not entirely sure what that means). It could also be Route.

pkg/controller/kuberneteseventsource/reconcile.go
```diff
@@ -0,0 +1,197 @@
...
+	if t.Status.Sinkable != nil {
+		return fmt.Sprintf("http://%s/", t.Status.Sinkable.DomainInternal), nil
+	}
+
+	// for now we will try again as a targetable.
+	if tryTargetable {
+		t := duckv1alpha1.Target{}
+		err = duck.FromUnstructured(obj, &t)
+		if err != nil {
+			// logger.Warnf("Failed to deserialize targetable: %s", zap.Error(err))
```
vaikas-google: remove?

pkg/controller/kuberneteseventsource/reconcile.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	obj, err := r.fetchObjectReference(ctx, source.Namespace, source.Spec.Sink)
+	if err != nil {
+		return "", err
+	}
+	t := duckv1alpha1.Sink{}
+	err = duck.FromUnstructured(obj, &t)
+	if err != nil {
+		// logger.Warnf("Failed to deserialize sink: %s", zap.Error(err))
```
vaikas-google: remove?

pkg/controller/kuberneteseventsource/kuberneteseventsource_controller_test.go
```diff
@@ -0,0 +1,87 @@
...
+	// Please modify the instance object by adding required fields and then remove the following if statement.
+	if apierrors.IsInvalid(err) {
+		t.Logf("failed to create object, got an invalid object error: %v", err)
+		return
+	}
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	defer c.Delete(context.TODO(), instance)
+	g.Eventually(requests, timeout).Should(gomega.Receive(gomega.Equal(expectedRequest)))
+
+	deploy := &appsv1.Deployment{}
```
vaikas-google: I don't see where the Deployment is being created nor where do we actually start watching it.

grantr: This is left over from the scaffold generated by kubebuilder. Creating a deployment isn't implemented or tested yet :smiley:

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
```diff
@@ -0,0 +1,120 @@
...
+type KubernetesEventSourceSpec struct {
+	// Sink is a reference to an object that will resolve to a domain name to use
+	// as the sink.
+	// +optional
+	Sink *corev1.ObjectReference `json:"sink,omitempty"`
+}
+
+const (
+	// KubernetesEventSourceConditionReady has status True when the
+	// source is ready to send events.
```
vaikas-google: nit: does this imply that SinkHas been provided and resolved?

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
```diff
@@ -0,0 +1,120 @@
...
+	KubernetesEventSourceConditionReady = duckv1alpha1.ConditionReady
+
+	// KubernetesEventSourceConditionSinkProvided has status True when the source
+	// has been configured with a sink target.
```
vaikas-google: Does this mean the ref has been resolved and all that. I wouldn't expect this to be true just by setting the Spec.SInk field.

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
```diff
@@ -0,0 +1,120 @@
...
+	KubernetesEventSourceConditionSinkProvided duckv1alpha1.ConditionType = "SinkProvided"
+)
+
+var kubernetesEventSourceCondSet = duckv1alpha1.NewLivingConditionSet()
+
+// KubernetesEventSourceStatus defines the observed state of the source.
+type KubernetesEventSourceStatus struct {
+	// Conditions holds the state of a source at a point in time.
+	// +optional
+	// +patchMergeKey=type
```
vaikas-google: What do these do? for ducks we've been using jsonpatch.

vaikas-google: or, did that change?

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
grantr: This is copied from ContainerSource, but I think the meaning is more `SinkResolved` than `SinkProvided`. @n3wscott can we rename this?

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
grantr: Yes. SinkProvided=True is a prerequisite for Ready=True.

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
grantr: I think these enable the Patch operation but I'm not sure. They're used in the corev1 apis [here](https://github.com/kubernetes/api/blob/6e4b5aa967ee9a03868e3b4edad98f27e54894c9/core/v1/types.go#L2066) and many other places.

grantr:
This is no longer WIP.

The KubernetesEventSource controller creates a ContainerSource object with the specified sink and an image url configured by the controller-manager env var `K8S_RA_ADAPTER` (stolen from @adamharwayne's #39).

The KES Ready condition is always the same as the ContainerSource Ready.

@vaikas-google @n3wscott The ContainerSource currently only supports passing the sink into the container via the `--sink` argument. For the existing k8s RA to work, we'll need to either update it to take `--sink` or update ContainerSource to also set an env var (I think the RA expects `TARGET`).

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
n3wscott: `SinkResolved` seems reasonable.

PR #27, opened 2018-10-24
using cloud events for the sample heartbeat

n3wscott:
## Proposed Changes

  * Use CloudEvents from knative/eventing for the heartbeat sample.

**Release Note**
```release-note
NONE
```

/assign @grantr


cmd/event_receiver/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+	"github.com/knative/eventing/pkg/event"
+	"log"
+	"net/http"
+	"time"
+)
+
+type Heartbeat struct {
```
matzew: so... not sure, but could this `struct` be moved to a common place? it's now around twice ... sure it's demo ... but not sure if that could be in `pkg/hearbeat`

cmd/heartbeats/main.go
```diff
@@ -50,39 +46,57 @@ func init() {
...
+		postMessage(sink, hb)
+		time.Sleep(period)
 	}
+}
 
-	for {
-		send()
-		time.Sleep(time.Duration(period) * time.Second)
+// Creates a CloudEvent Context for a given heartbeat.
+func cloudEventsContext(hb *Heartbeat) *event.EventContext {
```
matzew: doesn't this func need to use the `hb` argument ?

n3wscott:
Thanks for the conflicts @vaikas-google :D

cmd/heartbeats/main.go
```diff
@@ -50,39 +46,57 @@ func init() {
 
 func main() {
 	flag.Parse()
+	var period time.Duration
+	if p, err := strconv.Atoi(periodStr); err != nil {
```
grantr: What about using `flag.IntVar` to move this conversion into the flag library?

n3wscott: it can't because of the array of strings in the Container Source

Gopkg.lock
```diff
@@ -854,6 +862,8 @@
     "github.com/emicklei/go-restful",
     "github.com/google/go-cmp/cmp",
     "github.com/google/go-cmp/cmp/cmpopts",
+    "github.com/google/uuid",
+    "github.com/knative/eventing/pkg/event",
```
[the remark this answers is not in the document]
n3wscott: Cross that bridge when we get there...

cmd/event_receiver/main.go
n3wscott: I don't really want to because they are a demo and I was trying to keep them independent.

cmd/heartbeats/main.go
```diff
@@ -50,39 +46,57 @@ func init() {
...
-	hb = &Heartbeat{
+	hb := &Heartbeat{
 		Sequence: 0,
 		Label:    label,
 	}
 
-	period, err := strconv.Atoi(periodStr)
-	if err != nil {
-		period = 5
+	for {
```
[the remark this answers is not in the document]
n3wscott: Did it.

n3wscott:
PTAL @grantr

PR #32, opened 2018-10-25
adding a new makefile, like the old makefile.

n3wscott:
Fixes https://github.com/knative/eventing-sources/issues/23

## Proposed Changes

  * Add a makefile that does not do work directly, but delegates to scripts in the hack dir.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,35 @@
+# This file is needed by kubebuilder but all functionality should exist inside
+# the hack/ files.
+
+all: generate manifests test verify
+	
+# Run tests
+test: generate manifests verify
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,35 @@
+# This file is needed by kubebuilder but all functionality should exist inside
+# the hack/ files.
+
+all: generate manifests test verify
```
grantr: Is verify needed here, since generate and manifests just ran successfully?

Makefile
```diff
@@ -0,0 +1,35 @@
...
+	
+# Run tests
+test: generate manifests verify
```
grantr: Same verify comment as above

PR #37, opened 2018-10-25
Refactor sink helpers

grantr:
Moved sink helpers to controller/sinks package since they aren't generic.

Removed superfluous logging to eliminate the context argument and keep logging out of library functions.

Get namespace from ObjectRef instead of separate parameter. The ContainerSource reconciler always forces this to be the same as the source namespace, same as before.

Create a new function for setting the sink argument since I find several early returns easier to follow than several nested ifs.

Change the condition reason from "NotFound" to "Missing" when the Source has no sink ref, since the user action required is different. This should really be validated by the webhook but we don't have a
webhook yet.

Also fixed a few comment typos caused by function renames.

/cc @n3wscott


grantr:
@adamharwayne made a convincing argument that the `namespace` argument of `GetSinkURI` should remain because it's always required and burying it in the ObjectRef makes it easy to miss. I rolled that change back.

n3wscott:
/lgtm 

Thanks for the cleanup!!

PR #40, opened 2018-10-26
stdin is a file not a directory

grantr:
`make deploy` fails without this change, on Linux anyway.

/cc @n3wscott


n3wscott:
/lgtm

the command works on mac...

grantr:
Does it still work on mac with this change?

PR #43, opened 2018-10-29
Add support for manipulating Finalizers to the SDK

### LINKED ISSUE #41: Support Finalizers for sdk
/assign @adamharwayne

adamharwayne:
Fixes #41

## Proposed Changes

  * Add support for manipulating Finalizers to the SDK.

**Release Note**

```release-note
NONE
```


pkg/controller/sdk/reconciler.go
```diff
@@ -104,34 +101,49 @@ func (r *Reconciler) InjectConfig(c *rest.Config) error {
...
+	if !equality.Semantic.DeepEqual(oFinalizers, nFinalizers) {
+		return true
 	}
 
-	return true, nil
+	return false
 }
 
-func (r *Reconciler) updateStatus(ctx context.Context, request reconcile.Request, object runtime.Object) (runtime.Object, error) {
+func (r *Reconciler) update(ctx context.Context, request reconcile.Request, object runtime.Object) (runtime.Object, error) {
```
n3wscott: I would revert this and keep `updateStatus` and make a new `updateFinalizers` and `update`, have `update` call both those new methods. It will make testing easier.

pkg/controller/sdk/status_accessor.go
```diff
@@ -73,3 +73,56 @@ func (r *reflectedStatusAccessor) SetStatus(status interface{}) {
 		r.status.Set(reflect.ValueOf(status))
 	}
 }
+
+// FinalizersAccessor is the interface for a Resource that implements the getter and setting for
+// accessing its Finalizer set.
+// +k8s:deepcopy-gen=true
+type FinalizersAccessor interface {
```
n3wscott: This should move to it's own go file. Perhaps `finalizer_accessor.go`

pkg/controller/sdk/reconciler.go
```diff
@@ -67,24 +68,20 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	// Reconcile this copy of the Source and then write back any status
 	// updates regardless of whether the reconcile error out.
-	obj, err = r.provider.Reconciler.Reconcile(ctx, obj)
-	if err != nil {
-		logger.Warnf("Failed to reconcile %s: %v", r.provider.Parent.GetObjectKind(), err)
+	obj, reconcileErr := r.provider.Reconciler.Reconcile(ctx, obj)
```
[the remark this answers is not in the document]
adamharwayne: My reasoning is that this `err` is treated differently than most errors (at least most that I have seen in my limited time in Go). It is _not_ used right after being set. Instead its value needs to be retained until _after_ the update loop portion runs. I use a distinct name because I think it clues in the reader that this is not being handled like a standard error (namely that `err` can be overwritten).

pkg/controller/sdk/reconciler.go
```diff
@@ -67,24 +68,20 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-		// If we didn't change anything then don't call updateStatus.
-		// This is important because the copy we loaded from the informer's
-		// cache may be stale and we don't want to overwrite a prior update
-		// to status with this stale state.
-		return reconcile.Result{}, err
-	} else if _, err := r.updateStatus(ctx, request, obj); err != nil {
-		logger.Warnf("Failed to update %s status: %v", r.provider.Parent.GetObjectKind(), err)
-		return reconcile.Result{}, err
+	if needsUpdate := r.needsUpdate(ctx, original, obj); needsUpdate {
+		if _, updateErr := r.update(ctx, request, obj); updateErr != nil {
```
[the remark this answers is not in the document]
adamharwayne: Similar reasoning as above. We don't want the error from the reconcile call to be overwritten, so I don't want to use the same variable name. I know that due to being scoped in the if statement, this doesn't actually happen, but if both are called `err` and the ':' is removed from this line, then everything compiles but is logically wrong. I want to prevent that simple change from breaking things.

I suppose I really just want `reconcileErr` and can switch this back to `err` if you strongly prefer it.

pkg/controller/sdk/reconciler.go
```diff
@@ -67,24 +68,20 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
+			logger.Warnf("Failed to update %s: %v", r.provider.Parent.GetObjectKind(), updateErr)
```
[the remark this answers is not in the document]
adamharwayne: Done.

pkg/controller/sdk/status_accessor.go
adamharwayne: Done.

pkg/controller/sdk/reconciler.go
adamharwayne: I want there to be only a single call to `r.client.Update()`. Do you think it is worth pulling the 4 lines updating status and 4 lines updating finalizers into their own methods?

I disagree about making testing easier, because (where possible) we should be testing only the public interface, not the internal methods directly. And it is possible to test this without needing direct access to update.

pkg/controller/sdk/reconciler.go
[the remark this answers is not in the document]
n3wscott: I prefer, but not strongly... I will let it slide but if I ever pass over this code you know it is changing back :D

pkg/controller/sdk/reconciler.go
n3wscott: ok

n3wscott:
/lgtm

Looks like the branch is out of date.

pkg/controller/sdk/finalizers_accessor.go
```diff
@@ -0,0 +1,74 @@
...
+// FinalizersAccessor is the interface for a Resource that implements the getter and setting for
+// accessing its Finalizer set.
+// +k8s:deepcopy-gen=true
+type FinalizersAccessor interface {
+	GetFinalizers() interface{}
+	SetFinalizers(finalizers interface{})
+}
+
+// NewReflectedFinalizersAccessor uses reflection to return a FinalizersAccessor to access the field
+// called "Finalizers".
```
evankanderson: Mention this might return `nil`.

pkg/controller/sdk/finalizers_accessor.go
```diff
@@ -0,0 +1,74 @@
...
+func NewReflectedFinalizersAccessor(object interface{}) FinalizersAccessor {
+	objectValue := reflect.Indirect(reflect.ValueOf(object))
+
+	// If object is not a struct, don't even try to use it.
+	if objectValue.Kind() != reflect.Struct {
+		return nil
+	}
+
+	finalizerField := objectValue.FieldByName("Finalizers")
```
evankanderson: Should this dig through ObjectMeta as well, or no?

pkg/controller/sdk/finalizers_accessor.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"reflect"
+)
+
+// FinalizersAccessor is the interface for a Resource that implements the getter and setting for
+// accessing its Finalizer set.
+// +k8s:deepcopy-gen=true
+type FinalizersAccessor interface {
+	GetFinalizers() interface{}
+	SetFinalizers(finalizers interface{})
```
evankanderson: Should this be an `interface{}` or a stronger interface?

pkg/controller/sdk/finalizers_accessor.go
```diff
@@ -0,0 +1,74 @@
...
+package sdk
+
+import (
+	"reflect"
+)
+
+// FinalizersAccessor is the interface for a Resource that implements the getter and setting for
+// accessing its Finalizer set.
+// +k8s:deepcopy-gen=true
+type FinalizersAccessor interface {
```
vaikas-google: I thought finalizers were simply a list of strings?

pkg/controller/sdk/finalizers_accessor.go
```diff
@@ -0,0 +1,74 @@
...
+	GetFinalizers() interface{}
```
vaikas-google: or put another way, why not:
 GetFinalizers() []string
or what I've seen typically happen is using the Set abstraction for these.

vaikas-google: k8s.io/apimachinery/pkg/util/sets

pkg/controller/sdk/reconciler.go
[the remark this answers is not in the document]
evankanderson: Note that flattening everything to `err` here will break the code in a subtle way. Do we have a test that verifies the error from `r.provider.Reconciler.Reconcile(...)` is returned so that the tests will break if the error from 73 is dropped?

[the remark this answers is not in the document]
adamharwayne: I kept `reconcileErr` and changed this one back to `err` (as it follows the 'normal' Go error semantics of check immediately then throw away).

> Do we have a test that verifies the error from r.provider.Reconciler.Reconcile(...) is returned so that the tests will break if the error from 73 is dropped?

I don't think there are any tests for this SDK yet.

pkg/controller/sdk/finalizers_accessor.go
adamharwayne: Switched to using `sets.String`. It isn't useful for this change, but if we want to manipulate Finalizers in the future, it will be much easier.

pkg/controller/sdk/finalizers_accessor.go
adamharwayne: Switched to using `sets.String`. It isn't useful for this change, but if we want to manipulate Finalizers in the future, it will be much easier.

pkg/controller/sdk/finalizers_accessor.go
adamharwayne: Switched to using `sets.String`. It isn't useful for this change, but if we want to manipulate Finalizers in the future, it will be much easier.

pkg/controller/sdk/finalizers_accessor.go
adamharwayne: No. ObjectMeta is added to the struct via [embedding](https://github.com/knative/eventing-sources/blob/58dd271c37f4b085ccaf39ca61e6975f699330bc/pkg/apis/sources/v1alpha1/containersource_types.go#L131), so there is no ObjectMeta to dig through.

pkg/controller/sdk/finalizers_accessor.go
adamharwayne: Changed it to return (FinalizersAccessor, error) instead. If error == nil, then FinalizersAccessor != nil. I think this is standard Go style, do you want me to document it?

evankanderson: No, that seems fine. I think returning a `nil` interface is fine, too (and reasonable Go style), but documentation or explicit error helps to make it obvious that `nil` is a possible return value.

PR #52, opened 2018-10-30
K8s source controller updates

grantr:
## Proposed Changes

  * **Copy KES-owned ContainerSource conditions to KES.** This makes the status of the KES more obvious by hoisting every condition from the ContainerSource.
  * **Update the ContainerSource spec if needed.** If the Spec of the KES changes, the ContainerSource needs to be updated too. Use DeepEqual to only update when necessary.
  * **Add additional objects required for KES sample.** The KES sample also needs a channel and RBAC to read events. A future PR should allow the KES to take a service account to pass through to the
receive adapter pod.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/config/samples/sources_v1alpha1_kuberneteseventsource.yaml
+++ b/config/samples/sources_v1alpha1_kuberneteseventsource.yaml
@@ -8,4 +8,47 @@ spec:
   sink:
     apiVersion: eventing.knative.dev/v1alpha1
     kind: Channel
-    name: foo-channel
+    name: k8ssource
+    
+---
+  
+apiVersion: eventing.knative.dev/v1alpha1
+kind: Channel
+metadata:
+  name: k8ssource
+spec:
+  provisioner:
+    name: in-memory-channel
+    
+---
+  
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: event-watcher
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - get
+  - list
+  - watch
+  
+---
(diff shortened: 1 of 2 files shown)


config/samples/sources_v1alpha1_kuberneteseventsource.yaml
```diff
@@ -8,4 +8,47 @@ spec:
...
+  provisioner:
+    name: in-memory-channel
+    
+---
+  
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: event-watcher
```
n3wscott: how does event-watcher get passed to/through the source?

pkg/controller/kuberneteseventsource/reconcile.go
```diff
@@ -114,18 +114,17 @@ func (r *reconciler) reconcile(ctx context.Context, source *sourcesv1alpha1.Kube
...
-	if cs.Status.IsReady() {
-		source.Status.MarkReady()
-	} else {
-		reason := ""
-		message := ""
-		if ready := cs.Status.GetCondition(sourcesv1alpha1.ContainerConditionReady); ready != nil {
-			reason = ready.Reason
-			message = ready.Message
+	// Update ContainerSource spec if it's changed
+	expected := resources.MakeContainerSource(source, r.receiveAdapterImage)
```
n3wscott: nice. In time I would love to bubble this up to the sdk as the children reconciliation/update.

n3wscott:
Question answered:

>A future PR should allow the KES to take a service account to pass through to the
receive adapter pod.

/hold cancel

PR #60, opened 2018-11-01
Add ServiceAccountName to ContainerSource and KES

### LINKED ISSUE #49: Need ability to specify which service account to run *Sources as
## Expected Behavior
For example, in #48 the Receive Adapter which connects to the k8s API server, it needs permissions to be able to watch for events. Currently the Deployment created by KubernetesEventSource (and because it uses ContainerSource) runs as the default service account: 
  serviceAccount: default
  serviceAccountName: default

That means that the user needs to grant this account permissions to watch for k8s events which is undesirable. It would be better to be able to specify the Service Account that should be used so that it can be RBAC'd appropriately.

## Actual Behavior
Doesn't work unless you grant rights to default/default ServiceAccount

  1. Create KubernetesEventSource or ContainerSource and watch 
  2. Observe the Deployment / Pods created
  3. See they run as default service account.

## Additional Info
duck for namespaceable? :)

grantr:
Fixes #49

## Proposed Changes

  * Add `ServiceAccountName` to ContainerSource and pass it through to the Deployment.
  * Add `ServiceAccountName` to KES and pass it through to the ContainerSource.

**Release Note**

```release-note
ContainerSource now supports ServiceAccountName for setting the service account of
created Pods.

KubernetesEventSource now supports ServiceAccountName for setting the service
account of created Pods.
```

/cc @vaikas-google @n3wscott


pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
```diff
@@ -35,6 +35,11 @@ type KubernetesEventSourceSpec struct {
 	// Namespace that we watch kubernetes events in.
 	Namespace string `json:"namespace"`
 
+	// ServiceAccountName is the name of the ServiceAccount to use to run this
+	// source.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName"`
```
adamharwayne: omitempty

pkg/controller/containersource/resources/deployment.go
```diff
@@ -47,7 +47,6 @@ func MakeDeployment(org *appsv1.Deployment, args *ContainerArguments) *appsv1.De
 			Namespace:    args.Namespace,
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas: func() *int32 { var i int32 = 1; return &i }(),
```
adamharwayne: Was this intended to be removed?

If not I think the following is clearer.

```go
Replicas: int32(1)
```

grantr: It was intended to be removed.

`Spec.Replicas` is a pointer to an int to distinguish between unset and the valid value `0`. Go doesn't allow pointers to int literals so this function _thing_ was necessary to create an int pointer in one line.

I removed it because:
- It eliminates that inline function
- we always set `Replicas` to 1 which is the default for ReplicaSet anyway
- I think the correct behavior is to inherit the default number of replicas, whatever that happens to be (and it might not always be 1).

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
grantr: Done.

adamharwayne:
/lgtm

We should also add a doc with the 'standard' Source fields. I think so far they are:
- serviceAccountName
- sink

pkg/apis/sources/v1alpha1/containersource_types.go
```diff
@@ -49,6 +49,11 @@ type ContainerSourceSpec struct {
 	// +patchStrategy=merge
 	Env []corev1.EnvVar `json:"env,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
 
+	// ServiceAccountName is the name of the ServiceAccount to use to run this
+	// source.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
vaikas-google: since this is optional, if it's left out, does the right thing happen?

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
```diff
@@ -35,6 +35,11 @@ type KubernetesEventSourceSpec struct {
 	// Namespace that we watch kubernetes events in.
 	Namespace string `json:"namespace"`
 
+	// ServiceAccountName is the name of the ServiceAccount to use to run this
+	// source.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
vaikas-google: ditto.

pkg/apis/sources/v1alpha1/containersource_types.go
grantr: Yes. The empty ServiceAccountName value passes down to the Pod via the Deployment's PodSpec and is set to `default` by the apiserver.

pkg/apis/sources/v1alpha1/kuberneteseventsource_types.go
grantr: Yep (see above)

grantr:
> We should also add a doc with the 'standard' Source fields. I think so far they are:

Sure, where should this live? Is it part of this PR?

grantr:
Oops, shouldn't have merged master - didn't mean to remove lgtm

adamharwayne:
> > We should also add a doc with the 'standard' Source fields. I think so far they are:
> 
> Sure, where should this live? Is it part of this PR?

I'm not sure where in the repo it should live. But it should be linked from the 'make a new source' readme (which I assume is the main repo's readme for now).

PR #61, opened 2018-11-02
update knative/pkg and add prune options

n3wscott:
## Proposed Changes

  * update knative/pkg and add prune options

**Release Note**
```release-note
NONE
```


Gopkg.toml
```diff
@@ -18,16 +18,28 @@ required = [
 
 [prune]
   go-tests = true
+  unused-packages = true
```
n3wscott: we were not aggressive about prune, fixed

PR #63, opened 2018-11-02
do not respect crd status generation.

n3wscott:
## Proposed Changes

  * allow k8s >1.11 to apply source yaml.

**Release Note**
```release-note
Fix for applying source yaml to k8s > 1.11
```

### DIFF
--- a/config/crds/sources_v1alpha1_containersource.yaml
+++ b/config/crds/sources_v1alpha1_containersource.yaml
@@ -38,6 +38,8 @@ spec:
             image:
               minLength: 1
               type: string
+            serviceAccountName:
+              type: string
             sink:
               type: object
           type: object
--- a/config/crds/sources_v1alpha1_containersource.yaml
+++ b/config/crds/sources_v1alpha1_containersource.yaml
@@ -72,5 +72,5 @@ status:
   acceptedNames:
     kind: ""
     plural: ""
-  conditions: null
-  storedVersions: null
+  conditions: []
+  storedVersions: []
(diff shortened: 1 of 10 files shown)


config/crds/sources_v1alpha1_containersource.yaml
```diff
@@ -38,6 +38,8 @@ spec:
             image:
               minLength: 1
               type: string
+            serviceAccountName:
+              type: string
```
bbrowning: These serviceAccountName changes look unrelated to this specific fix - probably slipped in when generating the manifests based on some local changes to the container and kubernetes event source types?

grantr:
This was fixed upstream in https://github.com/kubernetes-sigs/controller-tools/pull/80. We're using a forked version of that dep to fix a different issue, so @n3wscott will try to get that patch into our fork. Hopefully that will make  this change unnecessary.

config/crds/sources_v1alpha1_containersource.yaml
n3wscott: someone forgot to run the gen command.

n3wscott: @grantr ^^

but I am not sure why the verify script did not catch this.

grantr: :hand: This should have gone into #60, but I'm also curious why the verify script didn't work.

bbrowning:
Thanks for the quick fix!

PR #69, opened 2018-11-05
Add a release.yaml default release

grantr:
Identical to `release-without-gcppubsub.yaml`, so now there are three variants:

- `release.yaml`
- `release-without-gcppubsub.yaml` (identical to `release.yaml`)
- `release-with-gcppubsub.yaml`

I don't have a strong opinion about `release-without-gcppubsub.yaml`. It seems like it might be useful, but it might just be confusing. We'll have a different release structure anyway once we break sources out into their own repos.

/cc @evankanderson @vaikas-google

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -26,6 +26,7 @@ readonly EVENTING_SOURCES_RELEASE_GCR
 
 # Yaml files to generate, and the source config dir for them.
 declare -A RELEASES
+RELEASES["release.yaml"]="config/default.yaml"
 RELEASES["release-without-gcppubsub.yaml"]="config/default.yaml"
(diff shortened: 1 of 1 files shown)


hack/release.sh
```diff
@@ -26,6 +26,7 @@ readonly EVENTING_SOURCES_RELEASE_GCR
 
 # Yaml files to generate, and the source config dir for them.
 declare -A RELEASES
+RELEASES["release.yaml"]="config/default.yaml"
 RELEASES["release-without-gcppubsub.yaml"]="config/default.yaml"
```
evankanderson:
```suggestion

```

PR #77, opened 2018-11-06
Adding status accessor tests in SDK.

n3wscott:
Adding tests to the SDK for the reconciler.

## Proposed Changes

  * Adding unit tests to status accessor helper function

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/sdk/status_accessor.go
+++ b/pkg/controller/sdk/status_accessor.go
@@ -17,6 +17,8 @@ limitations under the License.
 package sdk
 
 import (
+	"errors"
+	"fmt"
 	"reflect"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/controller/sdk/status_accessor_test.go
```diff
@@ -0,0 +1,108 @@
...
+			initial: v1alpha1.ContainerSourceStatus{},
+		},
+		"set status": {
+			obj:     &v1alpha1.ContainerSource{},
+			initial: v1alpha1.ContainerSourceStatus{},
+			update: v1alpha1.ContainerSourceStatus{
+				SinkURI: "just testing",
+			},
+		},
+		"set status with conditions": {
```
Harwayne: What is the difference between this test and "set status"?

n3wscott: this one has conditions

n3wscott: :D

Harwayne:
/lgtm

Is there a way to diff against an older commit? The rebase meant I had to reread the entire change, rather than just the ~5 lines that I think changed.

grantr:
> Is there a way to diff against an older commit?

Only if you know the commit id of the older commit. It might still exist in the repo, even if it's not referenced. This is why I prefer not to rebase when code is under review. 😺

pkg/controller/sdk/status_accessor.go
```diff
@@ -44,10 +46,10 @@ func NewReflectedStatusAccessor(object interface{}) StatusAccessor {
 		if _, ok := statusField.Interface().(interface{}); ok {
 			return &reflectedStatusAccessor{
 				status: statusField,
-			}
+			}, nil
```
grantr: Optional: I find code easier to read when the happy path is the final path. In this case, the happy path is the next to final path, and to understand that I need to understand that the condition is meant to be positive rather than negative.

The way the functions are named, it's weird for them to be a negative condition, so I understand why it was written this way. It's just not what I was expecting. :grin:

PR #85, opened 2018-11-08
Refactor the k8s events adapter, adding test 

n3wscott:
## Proposed Changes

  * [x] Move bulk of logic in the receive adapter out of the main package.
  * [x] Add tests to the new package.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/cmd/kuberneteseventsource/main.go
+++ b/cmd/kuberneteseventsource/main.go
@@ -17,177 +17,47 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
-	"fmt"
-	"io/ioutil"
 	"log"
-	"net/http"
-	"strings"
 
-	"github.com/golang/glog"
-	"github.com/knative/pkg/cloudevents"
-	"github.com/knative/pkg/signals"
-	corev1 "k8s.io/api/core/v1"
-	coreinformers "k8s.io/client-go/informers/core/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/clientcmd"
+	"github.com/knative/eventing-sources/pkg/adapter/kubernetesevents"
+	"go.uber.org/zap"
 )
 
-type EventWatcher struct {
-	target string
-}
+var (
+	sink      string
+	namespace string
+)
 
-func NewEventWatcher(target string) *EventWatcher {
-	return &EventWatcher{target: target}
-}
+func init() {
+	flag.StringVar(&sink, "sink", "", "uri to send events to")
+	flag.StringVar(&namespace, "namespace", "default", "namespace to watch events for")
 
-func (e *EventWatcher) updateEvent(old, new interface{}) {
-	e.addEvent(new)
-}
-
-func (e *EventWatcher) addEvent(new interface{}) {
-	event := new.(*corev1.Event)
-	log.Printf("GOT EVENT: %+v", event)
-	postMessage(e.target, event)
 }
 
 func main() {
-	namespace := flag.String("namespace", "", "namespace to watch events for")
-	target := flag.String("sink", "", "uri to send events to")
-
 	flag.Parse()
-
-	if namespace == nil || *namespace == "" {
-		log.Fatalf("No namespace given")
-	}
-
-	if target == nil || *target == "" {
-		log.Fatalf("No target given")
-	}
-
-	// set up signals so we handle the first shutdown signal gracefully
-	stopCh := signals.SetupSignalHandler()
-
-	log.Printf("Target is: %q", *target)
-	log.Printf("Namespace is: %q", *namespace)
-
-	cfg, err := clientcmd.BuildConfigFromFlags("", "")
+	ctx := context.Background()
+	logger, err := zap.NewProduction()
 	if err != nil {
-		log.Fatalf("Error building kubeconfig: %v", err)
+		log.Fatalf("Unable to create logger: %v", err)
 	}
 
-	kubeClient, err := kubernetes.NewForConfig(cfg)
-	if err != nil {
-		log.Fatalf("Error building kubernetes clientset: %v", err)
+	if namespace == "" {
+		logger.Fatal("No namespace provided")
 	}
 
-	log.Printf("Creating a new Event Watcher...")
-	watcher := NewEventWatcher(*target)
-
-	eventsInformer := coreinformers.NewFilteredEventInformer(
-		kubeClient, *namespace, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}, nil)
-
-	eventsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    watcher.addEvent,
-		UpdateFunc: watcher.updateEvent,
-	})
-
-	log.Printf("Starting eventsInformer...")
-	go eventsInformer.Run(stopCh)
-
-	log.Printf("Waiting for caches to sync...")
-	if ok := cache.WaitForCacheSync(stopCh, eventsInformer.HasSynced); !ok {
-		glog.Fatalf("Failed to wait for events cache to sync")
+	if sink == "" {
+		logger.Fatal("No sink provided")
 	}
-	log.Printf("Caches synced...")
-	<-stopCh
+
+	a := kubernetesevents.Adapter{
+		Namespace: namespace,
+		SinkURI:   sink,
+	}
+
+	a.Run(ctx)
+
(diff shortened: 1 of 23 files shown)


n3wscott:
Depends on https://github.com/knative/eventing-sources/pull/81

cmd/kuberneteseventsource/main.go
```diff
@@ -17,177 +17,47 @@ limitations under the License.
...
-		EventTime:          m.ObjectMeta.CreationTimestamp.Time,
+	a := kubernetesevents.Adapter{
+		Namespace: namespace,
+		SinkURI:   sink,
 	}
-}
 
-func postMessage(target string, m *corev1.Event) error {
-	ctx := cloudEventsContext(m)
+	a.Run(ctx)
```
grantr: Is the logger supposed to be added to this context? I don't see that it's ever set.

pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -0,0 +1,122 @@
...
+	Namespace string
+	// SinkURI is the URI messages will be forwarded on to.
+	SinkURI string
+
+	ctx context.Context
+}
+
+// Run starts the kubernetes receive adapter and sends events produced in the
+// configured namespace and send them to the sink uri provided.
+func (a *Adapter) Run(ctx context.Context) {
```
grantr: I recommend changing this signature to conform to the [`Runnable` interface](https://github.com/kubernetes-sigs/controller-runtime/blob/81f67a05853ea208098ffb64ad2862a8892771ba/pkg/manager/manager.go#L130-L135), i.e. `Start(<-chat struct{}) error`.

The benefit of this is that the global code in this function (setting up signal handling, exiting the process) can be handled entirely in `main`, making this function more testable and easier to use in combination with other things.

A minor benefit is that controller-runtime's Manager could be brought in to take care of things like creating a stop channel, setting up signal handlers, and creating a kubernetes client.

The context argument doesn't seem to be used at the moment (The context passed in from main is `context.Background()`) so I don't think it will be difficult to remove.

pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -0,0 +1,122 @@
...
+		AddFunc:    a.addEvent,
+		UpdateFunc: a.updateEvent,
+	})
+
+	logger.Debug("Starting eventsInformer...")
+	go eventsInformer.Run(stopCh)
+
+	logger.Debug("waiting for caches to sync...")
+	if ok := cache.WaitForCacheSync(stopCh, eventsInformer.HasSynced); !ok {
+		logger.Fatalf("Failed to wait for events cache to sync")
```
grantr: This function is package code so it shouldn't exit the process. It has no idea what else is running, or whether it's being exercised as part of a test.

pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -0,0 +1,122 @@
...
+		logger.Fatalf("error building kubernetes clientset: %v", zap.Error(err))
+	}
+
+	eventsInformer := coreinformers.NewFilteredEventInformer(
+		kubeClient, a.Namespace, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}, nil)
+
+	eventsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    a.addEvent,
+		UpdateFunc: a.updateEvent,
+	})
```
grantr: Optional, but this looks a lot like a controller. I wonder if we could save some lines (and maybe some bugs) by making this a standard controller-runtime reconciler. A subtlety is that we'd have to set the Manager's `Resync` option to `0` to disable resync so we don't get duplicate sends.

pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (a *Adapter) updateEvent(old, new interface{}) {
+	a.addEvent(new)
+}
+
+func (a *Adapter) addEvent(new interface{}) {
+	logger := logging.FromContext(a.ctx)
+	event := new.(*corev1.Event)
+	logger.Debugf("GOT EVENT: %+v", event)
```
grantr: Optional: Probably better to use the structured version `logger.Debug("GOT EVENT", zap.Any("event", event))` (and in other logger statements).

pkg/adapter/kubernetesevents/adapter_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestUpdateEvent_ServeHTTP(t *testing.T) {
+	testCases := map[string]struct {
+		sink         func(http.ResponseWriter, *http.Request)
```
grantr: I like this and the named functions. It's concise and clear.

pkg/adapter/kubernetesevents/adapter_test.go
```diff
@@ -0,0 +1,186 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			UID:               uid,
+			CreationTimestamp: metav1.Time{Time: now},
+		},
+		InvolvedObject: ref,
+	}
+
+	a.updateEvent(nil, event)
+
+	t.Log("should get here without crashing.")
```
grantr: I think this can be removed. If the test passes, this log message is never printed, and if it doesn't, there will be an error report of some kind and this message is superfluous.

pkg/adapter/kubernetesevents/adapter_test.go
```diff
@@ -0,0 +1,186 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			UID:               uid,
+			CreationTimestamp: metav1.Time{Time: now},
+		},
+		InvolvedObject: ref,
+	}
+
+	err := a.postMessage(event)
+
+	if err == nil {
```
grantr: Can we be a bit more selective on what error we're looking for? A different error might cause this to pass incorrectly.

pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -0,0 +1,122 @@
...
+	eventType = "dev.knative.k8s.event"
+)
+
+type Adapter struct {
+	// Namespace is the kubernetes namespace to watch for events from.
+	Namespace string
+	// SinkURI is the URI messages will be forwarded on to.
+	SinkURI string
+
+	ctx context.Context
```
grantr: This context is stored and used for the lifetime of the process, which seems like an invalid use of context. It might be simpler and more idiomatic to give the struct a Logger field, or alternately add a `NewAdapter` function that takes a logger and sets an unexported `logger` field using some transformation (`WithName`, etc).

pkg/adapter/kubernetesevents/adapter.go
n3wscott: oh good point!

pkg/adapter/kubernetesevents/adapter.go
n3wscott: I think this is a great idea. But perhaps it is a refactor after?

pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -0,0 +1,118 @@
...
+	eventsInformer := coreinformers.NewFilteredEventInformer(
+		kubeClient, a.Namespace, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}, nil)
+
+	eventsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    a.addEvent,
+		UpdateFunc: a.updateEvent,
+	})
+
+	logger.Debug("Starting eventsInformer...")
+	go eventsInformer.Run(stopCh)
```
matzew: @n3wscott I am getting now this:

```
E1112 17:00:33.932374       1 reflector.go:205] github.com/knative/eventing-sources/pkg/adapter/kubernetesevents/adapter.go:72: Failed to list *v1.Event: Get https://172.30.0.1:443/api/v1/namespaces/default/events?limit=500&resourceVersion=0: dial tcp 172.30.0.1:443: connect: connection refused
```

matzew: And, yes, that IP/port pair is used (`kubectl get svc -n default`):

```
kubernetes                          ClusterIP      172.30.0.1       <none>  ....
```

matzew: fixed / or better... gone .......

PR #88, opened 2018-11-08
Adding container source deployment conversion test.

n3wscott:
## Proposed Changes

  * Adding a test on the container source resources dir.

**Release Note**
```release-note
NONE
```

/assign @grantr


pkg/controller/containersource/resources/deployment_test.go
```diff
@@ -0,0 +1,93 @@
...
+					Key: "test2-secret",
+				},
+			},
+		}},
+		ServiceAccountName: "test-service-account",
+		SinkInArgs:         false,
+		Sink:               "test-sink",
+	}
+
+	want := `{"kind":"Deployment","apiVersion":"apps/v1","metadata":{"generateName":"test-name-","namespace":"test-namespace","creationTimestamp":null},"spec":{"selector":{"matchLabels":{"source":"test-name"}},"template":{"metadata":{"creationTimestamp":null,"labels":{"source":"test-name"},"annotations":{"sidecar.istio.io/inject":"true"}},"spec":{"containers":[{"name":"source","image":"test-image","args":["--test1=args1","--test2=args2","--sink=test-sink"],"env":[{"name":"test1","value":"arg1"},{"name":"test2","valueFrom":{"secretKeyRef":{"key":"test2-secret"}}},{"name":"SINK","value":"test-sink"}],"resources":{},"imagePullPolicy":"IfNotPresent"}],"serviceAccountName":"test-service-account"}},"strategy":{}},"status":{}}`
```
grantr: Why use a json string to compare instead of a Deployment object? Not saying it's wrong, but I'm curious about the rationale.

n3wscott: ONLY because I did not want to test creating the object by creating the same object. I would like ... just duplicate the code again.  So I wanted to compare a conversion of the result of the method

matzew: I like that approach

grantr: I understand that concern! IMO copy/pasting once is a reasonable approach since you have the opportunity to verify that it's correct manually when you write the test. In return, updating the test is easier for future maintainers, the intent is more clear (JSON marshaling is not the important part), and it's easier to reduce brittleness by ignoring irrelevant fields with `cmp.Diff` options.

n3wscott:
I am not sold yet that this is infact a good test....

Let me noodle

n3wscott:
/hold cancel

I changed how I am generating the assumed deployment and because it is easier to read, I found and fixed a bug !!

PTAL

pkg/controller/containersource/resources/deployment.go
```diff
@@ -79,3 +81,14 @@ func MakeDeployment(org *appsv1.Deployment, args *ContainerArguments) *appsv1.De
 
 	return deploy
 }
+
+func sinkArg(args *ContainerArguments) string {
+	if args.SinkInArgs {
+		for _, a := range args.Args {
+			if strings.HasPrefix(a, "--sink=") {
+				return strings.Replace(a, "--sink=", "", -1)
```
grantr: Good catch!

n3wscott: Weird when the tests help you find an issue!!

matzew:
Wat an awesome bot! :dog:

PR #90, opened 2018-11-09
Bump Serving to 0.2.1

bbrowning:
This bumps to the latest Serving release, as requested from a comment in #36 .


n3wscott:
@bbrowning is a scholar and a gentleman.

/lgtm

PR #92, opened 2018-11-09
Removing duplicated examples

matzew:
remove `gcppubsub` and `k8sevent` since those are already up-to-date in the docs repo; let's not duplicate them here

TODO

* right sample for `sources_v1alpha1_containersources.yaml`

Questions:

* Can we remove the samples in the `config/samples` folder, [here](https://github.com/knative/eventing-sources/tree/master/config/samples)?

Both, are in `docs` repo (Github is PR'd to the repo)


matzew:
@bbrowning @evankanderson thoughts ?

n3wscott:
/hold

In this case I think it makes more sense for these to stay in this repo. What do you think @evankanderson ?

n3wscott:
The reasoning is the eventing sources are knative sanctioned extensions to core knative. So knative/docs has core documentation, but this is a first party extension of knative and therefore it is documented locally. this is my opinion...

matzew:
I see knative/docs as a book. With chapters on eventing, and samples

* build
* serving
* eventing
  * sources
  * provisioners

Not sure it's that intuitive if samples are spread across different places. Just my $0.02 :smile:

grantr:
I mostly agree with @matzew. The reader of the knative docs probably doesn't care about what code lives where, they just want to be successful. And practically we can control the consistency of examples better if they all live in the same place, increasing the chance of success IMO (though on the other hand keeping docs up to date is harder when they're separate from the code).

However I think @n3wscott's points are relevant because the future we're hoping for is one with dozens of third-party sources. We will need to maintain the high quality and usability of the central knative/docs repo while also enabling the entire ecosystem to produce code and docs efficiently.

/cc @samodell 
Sam, do you have thoughts on where docs and examples should live?

samodell:
I think this is a topic that we should discuss in the Documentation Working Group. Long-term, figuring out how to curate a set of samples that makes sense and is feasible to maintain is incredibly important. On one level, I agree with @matzew that it makes sense for all documentation for the repos under knative/ to live in the docs repo, but if there are going to be a lot of "first-party extensions" living under knative/, it might quickly grow to a size that we can't reasonably maintain. 

For now, let's leave this documentation here, and I'll add this to the agenda for the Dec. 4th documentation working group meeting.

cc @grantr

evankanderson:
(Is conflicting, so /lgtm doesn't do much.)

My $0.02 given the current refactor of the repo into `contrib/` sources (Thanks @matzew !) is that we should probably have the `contrib/foo/README.md` act as a mini landing page for the source, documenting how it works, arguments/spec to the Source, install instructions/release links, etc. I'm not sure if we need a sample for each event source so much as reference documentation, unless there's a parlticularly cool demo (and I'm happy to leave the demos in docs' `eventing/samples` rather than next to the sources).

PR #93, opened 2018-11-09
Adding ContainerSource reconcile to happy test.

n3wscott:
## Proposed Changes

  * Reconcile to happy to test that the resource is ready true when the deployment is ready.

**Release Note**
```release-note
NONE
```


pkg/controller/containersource/reconcile.go
```diff
@@ -97,7 +97,7 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 				r.recorder.Eventf(source, corev1.EventTypeNormal, "DeploymentBlocked", "waiting for %v", err)
 				return object, err
 			}
-			r.recorder.Eventf(source, corev1.EventTypeNormal, "Deployed", "Created deployment %q", deploy.Name)
+			r.recorder.Eventf(source, corev1.EventTypeNormal, "Deployed", "Created deployment %q", args.Name)
```
grantr: Why not use the actual deployment name here? Won't it be different from `args.Name` (because it's generated)?

pkg/controller/containersource/reconcile.go
```diff
@@ -97,7 +97,7 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 				r.recorder.Eventf(source, corev1.EventTypeNormal, "DeploymentBlocked", "waiting for %v", err)
 				return object, err
 			}
-			r.recorder.Eventf(source, corev1.EventTypeNormal, "Deployed", "Created deployment %q", deploy.Name)
```
n3wscott: deploy.Name is empty here until the deployment controller reconciles once.

pkg/controller/containersource/reconcile.go
```diff
@@ -111,9 +111,10 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 	expected := resources.MakeDeployment(nil, args)
 	if !equality.Semantic.DeepEqual(deploy.Spec, expected.Spec) {
 		deploy.Spec = expected.Spec
-		if r.client.Update(ctx, deploy); err != nil {
-			return object, err
-		}
+		err := r.client.Update(ctx, deploy)
+		source.Status.MarkDeploying("DeployUpdated", "Updated deployment %s", deploy.Name)
```
grantr: Good catch! What should happen when `err != nil` here? I'm not sure whether status should change or not.

Seems like what will happen is:
- Mark status deploying
- Ready will become False
- Immediate reconcile again
- Attempt a second update

For a transient error, this is effectively the same as only updating status on success.

In the case of persistent errors, the Deployment never actually changes so the data plane is still technically working. On the other hand, the Deployment is out of date so from the perspective of the ContainerSource, there's more left to do and that seems like Ready=Unknown.

I guess the question is about the definition of Ready=True: Is it "events are plumbed", or "reconcile is complete"?

I see the [serving documentation about conditions](https://github.com/knative/serving/blob/22aacc4a25837fb1eac408cc8925707e557b0248/docs/spec/errors.md#knative-serving-condition-status) defines Ready as the latter.

So if we're conforming to the knative/serving conditions semantics (and we should), then **the status should be updated immediately after recognizing that reconciliation is required**. That's what's happening here, so this comment can be disregarded :grin:

pkg/controller/containersource/reconcile.go
```diff
@@ -111,9 +111,10 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
...
+		// Return after this update and reconcile again
+		return object, err
```
grantr: Good catch! I agree this should return, since we need to wait for the Deployment status to update anyway.

pkg/controller/containersource/reconcile_test.go
```diff
@@ -284,71 +282,231 @@ var testCases = []controllertesting.TestCase{
...
+				s := getContainerSource()
+				s.UID = containerSourceUID
+				return s
+			}(),
+			func() runtime.Object {
+				// TODO(n3wscott): this is very strange, I was not able to get
+				// the fake client to return the resources.MakeDeployment version
+				// back in the list call. I might have missed setting some special
+				// metadata? Converting an unstructured and setting the fields
+				// I care about did work.
```
grantr: I don't think mocking a response is necessary. After creating the deployment, the reconciler immediately returns, there is no situation when we need to reconcile a deployment that has just been created. From the perspective of the reconciler reacting to the Deployment, the Deployment already exists. It should be enough to add a Deployment object to the `InitialState` array, no mocks required.

pkg/controller/containersource/reconcile_test.go
```diff
@@ -284,71 +282,231 @@ var testCases = []controllertesting.TestCase{
...
+				s.UID = containerSourceUID
+				s.Status.InitializeConditions()
+				s.Status.MarkDeployed()
+				s.Status.MarkSink(targetURI)
+				return s
+			}(),
+		},
+		IgnoreTimes: true,
+	}, {
+		Name:       "xxxvalid containersource, sink, but deployment needs update",
```
grantr: Why the `xxx` here? Is this test unfinished?

n3wscott: ha! oops

pkg/controller/containersource/reconcile_test.go
```diff
@@ -50,6 +48,7 @@ const (
 	containerSourceName = "testcontainersource"
 	testNS              = "testnamespace"
 	containerSourceUID  = "2a2208d1-ce67-11e8-b3a3-42010a8a00af"
+	deployGeneratedName = "" //sad trombone
```
n3wscott: 🤣

n3wscott:
/hold cancel 

Had to fix a misunderstanding with generate name and the fake client.

pkg/controller/containersource/reconcile.go
evankanderson: This probably suggests that you need another condition to indicate whether the data plane is up to date, or you can treat this as `Ready = Unknown` until the Deployment matches the requested state.

pkg/controller/containersource/reconcile.go
```diff
@@ -111,9 +111,16 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
...
 		deploy.Spec = expected.Spec
-		if r.client.Update(ctx, deploy); err != nil {
-			return object, err
+		err := r.client.Update(ctx, deploy)
+		// if no error, update the status.
+		if err == nil {
+			source.Status.MarkDeploying("DeployUpdated", "Updated deployment %s", deploy.Name)
+		} else {
+			source.Status.MarkDeploying("DeployNeedsUpdate", "Attempting to update deployment %s", deploy.Name)
+			r.recorder.Eventf(source, corev1.EventTypeWarning, "DeployUpdated", "Failed to update deployment %q", deploy.Name)
```
grantr: :+1:

PR #95, opened 2018-11-09
Adding a test for gcp resource.

n3wscott:
## Proposed Changes

  * Test resource generation for pubsub source
  * Move ownership call to controller to reduce complexity in the Resource code area.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/gcppubsub/reconcile.go
+++ b/pkg/controller/gcppubsub/reconcile.go
@@ -19,6 +19,7 @@ package gcppubsub
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/knative/eventing-sources/pkg/controller/gcppubsub/resources"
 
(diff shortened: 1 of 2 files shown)


n3wscott:
/assign @grantr 
/cc @Harwayne 

Adam I did some refactoring, check it out.

pkg/controller/gcppubsub/resources/receive_adapter.go
```diff
@@ -23,29 +23,29 @@ import (
...
+	credsVolume    = "google-cloud-key"
+	credsMountPath = "/var/secrets/google"
+)
+
 // MakeReceiveAdapter generates (but does not insert into K8s) the Receive Adapter Deployment for
 // GCP PubSub Sources.
-func MakeReceiveAdapter(args ReceiveAdapterArgs) (*v1.Deployment, error) {
-	credsVolume := "google-cloud-key"
-	credsMountPath := "/var/secrets/google"
+func MakeReceiveAdapter(args *ReceiveAdapterArgs) *v1.Deployment {
```
grantr: :+1: Consolidating error handling in the caller seems like a readability improvement, and makes this function more testable.

Harwayne: I actually disagree. I dislike this refactor because it makes this function not create a usable `Deployment`. The caller has the add the owner reference themselves (which is not documented). If they don't, then I think the `Deployment` will leak after the `GcpPubSubSource` is deleted. 

It also opens the potential for a future bug (although very unlikely) where the source passed through args (line 166) is different than the one added as an owner (line 171).

PR #97, opened 2018-11-12
Remove manual pruning of knative/test-infra

grantr:
This was incorrectly removing `dep-collector`, and wasn't really necessary in the first place according to @mattmoor.

/cc @mattmoor @adrcunha


adrcunha:
You're now vendoring the whole test-infra, which is unnecessary. The packages need to be adjusted, just like in knative/serving#2420

grantr:
Thanks @adrcunha for noticing! I'll create a followup PR to fix that.

PR #99, opened 2018-11-13
Print the string instead of the pointer

grantr:
Fixes a compile error complaining about using a string directive on a pointer:

```
cmd/githubsource/main.go:57: Printf format %q has arg sink of wrong type *string 
```

I'm not sure why this doesn't come up in builds. Maybe it's specific to my go version?

```
$ go version             
go version go1.11 linux/amd64
```

/cc @bbrowning

### DIFF
--- a/cmd/githubsource/main.go
+++ b/cmd/githubsource/main.go
@@ -54,7 +54,7 @@ func main() {
 		log.Fatalf("No secret token given")
 	}
 
-	log.Printf("Sink is: %q", sink)
+	log.Printf("Sink is: %q", *sink)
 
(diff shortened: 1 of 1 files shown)


grantr:
@adrcunha what version of Go does Prow use?

bbrowning:
It looks like I need to update to go 1.11 instead of 1.10. Sorry about that.

/lgtm

adrcunha:
The go version is displayed in the test logs

grantr:
Thanks for pointing that out @adrcunha. I think we should, especially if 1.11 is more strict about what code is valid.

PR #107, opened 2018-11-14
Revert "Update the controller-runtime dependency."

Harwayne:
Reverts knative/eventing-sources#105

This change appears to make the ContainerSource never declare itself as ready. Events still flow on the data plane, but the Object is never marked as ready.


Harwayne:
Seems to be caused by an unrelated problem in `ContainerSource`.

PR #109, opened 2018-11-14
[WIP] Default KubernetesEventSource namespace

grantr:
If the Namespace field is empty, default to the same namespace as the source.

This is WIP because I'd like to get unit tests for KES in first.

## Proposed Changes

  * Default KubernetesEventSource watched namespace to same namespace as source

**Release Note**

```release-note
KubernetesEventSource will now watch events in the same namespace as the source by default.
```

/cc @vaikas-google

### DIFF
--- a/pkg/controller/kuberneteseventsource/resources/containersource.go
+++ b/pkg/controller/kuberneteseventsource/resources/containersource.go
@@ -26,6 +26,11 @@ import (
 // MakeContainerSource generates, but does not create, a ContainerSource for the
 // given KubernetesEventSource.
 func MakeContainerSource(source *sourcesv1alpha1.KubernetesEventSource, receiveAdapterImage string) *sourcesv1alpha1.ContainerSource {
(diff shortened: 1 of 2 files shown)


pkg/controller/kuberneteseventsource/resources/containersource.go
```diff
@@ -26,14 +26,19 @@ import (
 // MakeContainerSource generates, but does not create, a ContainerSource for the
 // given KubernetesEventSource.
 func MakeContainerSource(source *sourcesv1alpha1.KubernetesEventSource, receiveAdapterImage string) *sourcesv1alpha1.ContainerSource {
+	namespace := source.Spec.Namespace
```
Harwayne: My preference is to do this in a webhook, rather than in the controller itself. That way the applied default is easily visible to the user (although in this case, it is simple enough that I think everyone would understand what is happening).

PR #110, opened 2018-11-14
Use Status subresource

grantr:
Now that GKE has 1.11 available we can use the status subresource for CRDs. This eliminates the need to manipulate `Spec.Generation`, which is good because eventing-sources doesn't do that yet.

## Proposed Changes

  * Add status subresource to all source CRDs
  * Use status subresource to update KubernetesEventSource status

**Release Note**

```release-note
CRDs now enable the status subresource, and controllers use it to update status when possible.
```

### DIFF
--- a/config/crds/sources_v1alpha1_containersource.yaml
+++ b/config/crds/sources_v1alpha1_containersource.yaml
@@ -16,6 +16,8 @@ spec:
     kind: ContainerSource
     plural: containersources
   scope: Namespaced
+  subresources:
+    status: {}
   validation:
     openAPIV3Schema:
       properties:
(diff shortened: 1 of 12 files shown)


grantr:
This may have caused test failures in eventing. The evidence I used to confirm that GKE supported the status subresource was https://github.com/knative/serving/pull/786, which may not have been quite as authoritative as it should have been :smile: @dprotaso do you know the state of GKE's support for CRD subresources?

grantr:
After some more research I think the feature is actually enabled, so false alarm there. More details in #114.

dprotaso:
@grantr It's beta in 1.11 so it's on by default. Thus you can 'enable' the status subresource on your CRDs

PR #123, opened 2018-11-20
Refactor k8s source to use the sdk.

n3wscott:
## Proposed Changes

  * [x] Use sdk for k8s source.
  * [x] Add tests
 * Delete was not being checked in reconcile.
 * The first time you create a container source, it has no status so it would deep copy erase all the k8s source status conditions
 * Moved it to best practice for conditionals.

**Release Note**
```release-note
NONE
```

### DIFF
--- a/pkg/controller/kuberneteseventsource/provider.go
+++ b/pkg/controller/kuberneteseventsource/provider.go
@@ -18,85 +18,37 @@ package kuberneteseventsource
 
 import (
 	"fmt"
+	"github.com/knative/eventing-sources/pkg/controller/sdk"
 	"os"
 
 	sourcesv1alpha1 "github.com/knative/eventing-sources/pkg/apis/sources/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/record"
-	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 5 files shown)


pkg/controller/kuberneteseventsource/reconcile_test.go
```diff
@@ -0,0 +1,379 @@
...
+				s.UID = sourceUID
+				return s
+			}(),
+			func() runtime.Object {
+				u := getContainerSource_isReady()
+				u.SetOwnerReferences(getOwnerReferences())
+				return u
+			}(),
+		},
+		ReconcileKey: fmt.Sprintf("%s/%s", testNS, sourceName),
```
Harwayne: I recommend removing `ReconcileKey` and `IgnoreTimes` out of every test case and just set them in the test itself, because every test case uses the same values for those two fields.

pkg/controller/kuberneteseventsource/provider.go
```diff
@@ -18,85 +18,37 @@ package kuberneteseventsource
...
-	scheme              *runtime.Scheme
-	receiveAdapterImage string
-}
-
-var _ reconcile.Reconciler = &reconciler{}
-
-// Add creates a new KubernetesEventSource Controller and adds it to the Manager
-// with default RBAC. The Manager will set fields on the Controller and Start it
-// when the Manager is Started.
+// Add creates a new GitHubSource Controller and adds it to the
```
Harwayne: KubernetesSource

matzew:
@n3wscott what's the status here ?

Issue #55, opened 2018-11-01
inline k8s corev1 Container over ContainerSourceSpec


rootfs:
## Expected Behavior
[ContainerSourceSpec](https://github.com/knative/eventing-sources/blob/master/pkg/apis/sources/v1alpha1/containersource_types.go#L37) has some but not all k8s corev1 Container. If k8s corev1 Container is inlined here, we then don't need to hardcode the pull policy [here](https://github.com/knative/eventing-sources/blob/master/pkg/controller/containersource/resources/deployment.go#L72) and will make is easier for sources that base on ContainerSource.

## Steps to Reproduce the Problem

  1.
  2.
  3.

## Additional Info

@n3wscott @grantr

n3wscott:
Perhaps container source should bump out and have a the container template?

rootfs:
if that is not a big inconvenience, please get a pod template

n3wscott:
I think it is a good idea. @evankanderson and @grantr were looking at this a bit too

sebgoa:
+1 to this.

We need to be able to use a full Pod spec in the Container source manifest.

Being able to use secrets as env variables is one good use-case.

sebgoa:
@tzununbekov let's discuss this one, I think you could help here

vaikas-google:
Makes sense to add the pod template. however, you should be able to use env secrets today as is. not disputing the need for pods.

sebgoa:
confirmed env configured via secrets do work. However volumes mounts don't. 

If you create a container source for Google cloud for instance, you need to set GOOGLE_APPLICATION_CREDENTIALS to a file path where you mount the service account secret.

Therefore today we cannot write a Container source for Google services.

vaikas-google:
I'm closing this because this has been fixed by using the podtemplatespec. Re-open if there's something else that's needed.
https://github.com/knative/eventing/pull/1321

PR #124, opened 2018-11-20
AWS SQS source

srvaroa:
## Proposed Changes

  * Provide basic suport for an AWS SQS source.


srvaroa:
I have been able to run this successfully. I still want to give a pass to the READMEs and cleanup code but it should be in a decent state to get some feedback.

config/awssqs/istio_egress.yaml
```diff
@@ -0,0 +1,14 @@
+# Needed by the AWS SQS Source to communicate with AWS SQS.
+
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: awssqs-bus-ext
+spec:
+  hosts:
+  - "*.amazonaws.com"
```
srvaroa: This is actually not working and I had to use the FQDN (sqs.eu-west-1.amazonaws.com in my case) to make it work.

rootfs:
@srvaroa thanks for mentioning :) 

I am adopting container source for my new event source [here](https://github.com/ceph/rgw-pubsub-api/blob/master/go/examples/knative-eventing-source/container-source/main.go). It is quite similar to SQS. Except secret volume (which is to be addressed in #55, container source is ready for its prime time.

cc @grantr @n3wscott

config/awssqs/istio_egress.yaml
srvaroa: Thanks @rootfs. It looks like a false alarm.  The istio-proxy container takes longer to start than the receive-adapter so you can see this in the logs:

> default/awssqs-devhose-source-vzrsc-765c547b68-gmp7m[receive-adapter]: {"level":"info","ts":1542735771.9919236,"caller":"awssqs_receive_adapter/main.go:68","msg":"Starting AWS SQS Receive Adapter. %v","adapter":{"Region":"eu-west-1","QueueUrl":"...","SinkURI":"http://qux-1-channel.default.svc.cluster.local/","CredsFile":"/var/secrets/aws/credentials"}}
> default/awssqs-test-source-vzrsc-765c547b68-gmp7m[receive-adapter]: {"level":"info","ts":1542735771.9922512,"logger":"fallback","caller":"awssqs/adapter.go:63","msg":"Starting with config: {adapter 22 0  0xc4200cedc0}"}
> default/awssqs-test-source-vzrsc-765c547b68-gmp7m[receive-adapter]: {"level":"warn","ts":1542735772.3487105,"logger":"fallback","caller":"awssqs/adapter.go:77","msg":"Failed to poll from SQS queue{error 25 0  RequestError: send request failed\ncaused by: Post https://sqs.eu-west-1.amazonaws.com/: dial tcp 52.95.114.128:443: connect: connection refused}"}

But the istio-proxy does initialize correctly after a while

> default/awssqs-devhose-source-vzrsc-765c547b68-gmp7m[istio-proxy]: [2018-11-20 17:42:53.579][19][info][upstream] external/envoy/source/common/upstream/cluster_manager_impl.cc:494] add/update cluster outbound|80||message-dumper-00001-service.default.svc.cluster.local during init
> default/awssqs-devhose-source-vzrsc-765c547b68-gmp7m[istio-proxy]: [2018-11-20 17:42:53.579][19][info][upstream] external/envoy/source/common/upstream/cluster_manager_impl.cc:494] add/update cluster outbound|443||controller-manager-service.knative-sources.svc.cluster.local during init
> default/awssqs-devhose-source-vzrsc-765c547b68-gmp7m[istio-proxy]: [2018-11-20 17:42:53.580][19][info][upstream] external/envoy/source/common/upstream/cluster_manager_impl.cc:494] add/update cluster outbound|443||*.amazonaws.com during init

At this point, the receive adapter starts to poll successfully.

In my tests I had a plain exit if there was a poll error (I'm reading from a queue with a lot of messages so it was hard to read the logs while debugging.) I didn't realize that the sleep I added on error actually masks this behaviour.

cmd/awssqs_receive_adapter/main.go
```diff
@@ -0,0 +1,72 @@
...
+	"log"
+	"os"
+
+	"go.uber.org/zap"
+
+	"golang.org/x/net/context"
+)
+
+const (
+	// Sink for messages.
```
Harwayne: I think the style is to have the first word be the variable name itself.

```golang
// envSinkURI is the sink for messages.
```

cmd/awssqs_receive_adapter/main.go
```diff
@@ -0,0 +1,72 @@
...
+	envSinkURI = "SINK_URI"
+
+	// envRegion is the AWS region for the given SQS queue
+	envRegion = "AWS_REGION"
+
+	// envUrl is the URL of the SQS queue to consume messages from.
```
Harwayne: envUrl -> envQueueUrl

cmd/awssqs_receive_adapter/main.go
```diff
@@ -0,0 +1,72 @@
...
+	envQueueUrl = "AWS_SQS_URL"
```
Harwayne: Url -> URL

cmd/awssqs_receive_adapter/main.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	adapter := &awssqs.Adapter{
+		Region:    getRequiredEnv(envRegion),
+		QueueUrl:  getRequiredEnv(envQueueUrl),
+		SinkURI:   getRequiredEnv(envSinkURI),
+		CredsFile: getRequiredEnv(envCredsFile),
+	}
+
+	logger.Info("Starting AWS SQS Receive Adapter. %v", zap.Reflect("adapter", adapter))
```
Harwayne: I don't think you want the `%v`. Just have the second parameter is enough.

Harwayne: `zap.Reflect` ->`zap.Any`

pkg/adapter/awssqs/adapter.go
```diff
@@ -0,0 +1,188 @@
...
+	"net/http"
+	"strconv"
+	"time"
+)
+
+const (
+	eventType = "aws.sqs.message"
+)
+
+// Adapter implements the AWS SQS adapter to deliver SQS messages from a
```
Harwayne: Extra 'a'

pkg/adapter/awssqs/adapter.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (a *Adapter) Start(ctx context.Context) error {
+
+	logger := logging.FromContext(ctx)
+
+	stopCh := signals.SetupSignalHandler()
+	cctx, cancel := context.WithCancel(ctx)
+
+	logger.Info("Starting with config: ", zap.Reflect("adapter", a))
```
Harwayne: `zap.Reflect` -> `zap.Any`

pkg/adapter/awssqs/adapter.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// CredsFile is the full path of the AWS credentials file
+	CredsFile string
+}
+
+func (a *Adapter) Start(ctx context.Context) error {
+
+	logger := logging.FromContext(ctx)
+
+	stopCh := signals.SetupSignalHandler()
```
Harwayne: What is the purpose in this stop channel? Normally they get passed in, so something on the outside can stop this function.

pkg/adapter/awssqs/adapter.go
```diff
@@ -0,0 +1,188 @@
...
+	}))
+
+	q := sqs.New(sess)
+
+	go func() {
+		for {
+			messages, err := poll(cctx, q, a.QueueUrl, 10)
+			if err != nil {
+				logger.Warn("Failed to poll from SQS queue", zap.Error(err))
+				time.Sleep(2 * time.Second)
```
Harwayne: If this timeout becomes a member of the Adapter struct (defaulted to 2 seconds), it may make unit testing easier.

pkg/adapter/awssqs/adapter.go
```diff
@@ -0,0 +1,188 @@
...
+
+	sess := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigDisable,
+		Config:            aws.Config{Region: aws.String(a.Region)},
+		SharedConfigFiles: []string{a.CredsFile},
+	}))
+
+	q := sqs.New(sess)
+
+	go func() {
```
Harwayne: I don't think this function ever exits (even if stopCh is closed or cancel is called).

pkg/apis/sources/v1alpha1/aws_sqs_types.go
```diff
@@ -0,0 +1,157 @@
...
+// Check that AwsSqsSource can be validated and can be defaulted.
+var _ runtime.Object = (*AwsSqsSource)(nil)
+
+// Check that AwsSqsSource implements the Conditions duck type.
+var _ = duck.VerifyType(&AwsSqsSource{}, &duckv1alpha1.Conditions{})
+
+// AwsSqsSourceSpec defines the desired state of the source.
+type AwsSqsSourceSpec struct {
+	// Url of the SQS queue that we will poll from.
+	QueueUrl string `json:"queueUrl"`
```
Harwayne: `QueueUrl` -> `QueueURL`

pkg/controller/awssqs/reconcile.go
```diff
@@ -0,0 +1,194 @@
...
+		logger.Error("could not find AwsSqs source", zap.Any("object", object))
+		return object, nil
+	}
+
+	// This Source attempts to reconcile three things.
+	// 1. Determine the sink's URI.
+	//     - Nothing to delete.
+	// 2. Create a receive adapter in the form of a Deployment.
+	//     - Will be garbage collected by K8s when this AwsSqsSource is deleted.
+	// 3. Register that receive adapter as a Pull endpoint for the specified AWS SQS queue.
```
Harwayne: This doesn't seem to be happening. If it's not, then I think we don't need the finalizer.

pkg/controller/awssqs/reconcile.go
```diff
@@ -0,0 +1,194 @@
...
+	sinkURI, err := sinks.GetSinkURI(r.dynamicClient, src.Spec.Sink, src.Namespace)
+	if err != nil {
+		src.Status.MarkNoSink("NotFound", "")
+		return src, err
+	}
+	src.Status.MarkSink(sinkURI)
+
+	_, err = r.createReceiveAdapter(ctx, src, sinkURI)
+	if err != nil {
+		logger.Error("Unable to create the receive adapter", zap.Error(err))
```
Harwayne:
```golang
return src, err
```

pkg/controller/awssqs/reconcile.go
```diff
@@ -0,0 +1,194 @@
...
+
+func getLabels(src *v1alpha1.AwsSqsSource) map[string]string {
+	return map[string]string{
+		"knative-eventing-source":      controllerAgentName,
+		"knative-eventing-source-name": src.Name,
+	}
+}
+
+func generateSubName(src *v1alpha1.AwsSqsSource) string {
+	return fmt.Sprintf("knative-eventing-%s-%s-%s", src.Namespace, src.Name, src.UID)
```
Harwayne: I don't know if this is possible, but if we can, we should use a separator that can't be in names or namespaces, e.g. underscore rather than dash.

samples/awssqs_source/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. Replace the place holders in `awssqs-source.yaml`.
+    - `QUEUE_URL` should be replaced with your AWS SQS queue URL.
+    - `qux-1` should be replaced with the name of the `Channel` you want
+      messages sent to. If you deployed an unaltered `channel.yaml`
+      then you can leave it as `qux-1`.
+    - `awsCredsSecret` should be replaced if you are using a non-default
+      secret or key name for the receive adapter's credentials.
+
+1. Deploy `awssqs-source.yaml`.
```
Harwayne: I've recently started giving the command using sed for the replacements, which I think makes it easier. So any early step would be `export QUEUE_URL=REPLACE_ME` and `export AWS_REGION=REPLACE_ME`, then this step is

```shell
sed "s!QUEUE_URL!$QUEUE_URL!g" samples/awssqs-source/awssqs-source.yaml | sed "s/AWS_REGION/$AWS_REGION/g" | kubectl -n default apply -f -
```

samples/awssqs_source/README.md
```diff
@@ -0,0 +1,95 @@
...
+1. If the deployed `AwsSqsSource` is pointing at a `Channel` other than `qux-1`, modify `subscriber.yaml` by replacing `qux-1` with that `Channel`'s name.
+1. Deploy `subscriber.yaml`.
+
+    ```shell
+    ko -n default apply -f samples/awssqs-source/subscriber.yaml
+    ```
+
+### Publish
+
+Publish messages to your AWS SQS queue.
```
Harwayne: Mention that the queue URL should be changed (or if you use the shell variable suggested above, then $QUEUE_URL).

samples/awssqs_source/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. Deploy the `AwsSqsSource` controller as part of eventing-source's controller.
+
+    ```shell
+    ko -n default apply -f config/default-awssqs.yaml
+    ```
+
+    - Note that if the `Source` Service Account secret is in a non-default location, you will need to update the YAML first.
+
+1. Replace the place holders in `awssqs-source.yaml`.
```
Harwayne: Also the AWS_REGION and QUEUE_NAME.

pkg/adapter/awssqs/adapter.go
srvaroa: I copied this from @rootfs 's first attempt.  My understanding is that the SetupSignalHandler returns a channel that will emit on SIGTERM and SIGINT (see https://github.com/kubernetes/apiserver/blob/master/pkg/server/signal.go)

I'll leave it for now but let me know if that's incorrect (or the stop should be implemented in other ways later).

pkg/controller/awssqs/reconcile.go
srvaroa: You're right, leftover from the gcp pubsub code.

pkg/adapter/awssqs/adapter.go
Harwayne: Was the fix pushed?

pkg/adapter/awssqs/adapter.go
Harwayne: Sorry, my head must have been in `eventing` when I wrote this comment, where stop channels are almost always passed in. 

I still have a preference for the stop channel to be defined in `main()` and then passing it in, but that doesn't appear to be a pattern in `eventing-sources` yet, so I won't force it.

pkg/adapter/awssqs/adapter.go
srvaroa: Yes, https://github.com/knative/eventing-sources/pull/124/commits/6ca19c1c24528c28382fcf9f0fd7fe19d7c38583

Harwayne: This function (lines 80-109) never exits. The break on line 85 breaks out of the [select statement](https://golang.org/ref/spec#Break_statements) (I think, I'm also pretty new to go), but that just causes the for loop to run again. Because nothing in the function returns and nothing breaks out of the for loop, I think it will run forever.

srvaroa: Oh, I did not know break also applied to selects. I'll fix that. Thanks for the catch.

srvaroa:
@Harwayne I applied your other suggestion re. having stopCh parametrized to the start method in the adapter (see https://github.com/knative/eventing-sources/pull/124/commits/6cec8fc266981cfb919acec26d701bc49fd4dce3) - I couldn't reply on that thread

pkg/apis/sources/v1alpha1/aws_sqs_types.go
```diff
@@ -0,0 +1,157 @@
...
+// Check that AwsSqsSource can be validated and can be defaulted.
+var _ runtime.Object = (*AwsSqsSource)(nil)
+
+// Check that AwsSqsSource implements the Conditions duck type.
+var _ = duck.VerifyType(&AwsSqsSource{}, &duckv1alpha1.Conditions{})
+
+// AwsSqsSourceSpec defines the desired state of the source.
+type AwsSqsSourceSpec struct {
+	// QueueUrl of the SQS queue that we will poll from.
+	QueueUrl string `json:"queueUrl"`
```
whynowy: 1. Better to support both queue URL and queue ARN,
2. Region can be found either URL or ARN, does not need a specific configuration.

whynowy:
> @srvaroa - Since it's AWS SQS event source, I would assume people use SQS are running their K8s clusters on AWS EC2 instances. From within the cluster to access AWS resources (like SQS), there's another way to do it with kiam (https://github.com/uswitch/kiam) or kube2iam instead of using a long lived credential stored in k8s secret, I implemented the SQS event source by using kiam (https://github.com/whynowy/eventing-sources/commits/develop), I think it would be better to support both ways, what do you think?

As I know, a lot of people are using kiam to do delegated access to AWS resources, which is much more secure than using a stored credential. The approach of using secret is not going to be approved by the security team in my company.

pkg/apis/sources/v1alpha1/aws_sqs_types.go
srvaroa: Agree re. region. I'll fix that.

Re. ARN/URL I agree although for practical purposes the URL seems a better default choice given that the AWS API (https://docs.aws.amazon.com/sdk-for-go/api/service/sqs/#ReceiveMessageInput) takes URLs, so having ARN as an input would require extra transformation.

I'm happy to add it, although in the interest of keeping the PR's size contained I'd do that in a later PR. What do you think?

srvaroa:
Hi @whynowy thanks for the review. I agree that also having support for kiam would be good. Since it doesn't replace the secret-based approach (which covers also non AWS environments) I would add that support in a separate PR after this one is in (happy to do it myself or help out). WDYT? (cc @Harwayne @vaikas-google)

Harwayne:
> Hi @whynowy thanks for the review. I agree that also having support for kiam would be good. Since it doesn't replace the secret-based approach (which covers also non AWS environments) I would add that support in a separate PR after this one is in (happy to do it myself or help out). WDYT? (cc @Harwayne @vaikas-google)

I don't know enough about SQS to have an informed opinion. In general, I think this PR is looking pretty good and can get in soon, so my preference is to add the kiam support in a subsequent PR, rather than slow this one down. But, as I said, I don't know how important kiam support is.

srvaroa:
@Harwayne IAM is a better solution, but valid only if you're running your k8s cluster in AWS. The solution proposed here based on secrets would work regardless of where the k8s cluster runs (AWS included - I went down this route because I'm using GKE + and minikube for dev). The downside is that  it's less secure (you're spreading secrets around, if an attacker gets access to the k8s secrets that store the values they could use them to consume from the queue.) 

I think kiam support is important for the reasons @whynowy points out so I think it's a very desirable addition. I can work on this shortly after.

pkg/adapter/awssqs/adapter.go
```diff
@@ -0,0 +1,211 @@
...
+// emits an element.  The
+func (a *Adapter) pollLoop(ctx context.Context, q *sqs.SQS, stopCh <-chan struct{}) error {
+
+	logger := logging.FromContext(ctx)
+	cctx, cancel := context.WithCancel(ctx)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
```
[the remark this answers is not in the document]
srvaroa: I was being conservative with the pauses, but you're right that we can err on the side of simplicity. I've lowered the timeout to 3 to give us narrower loops, and removed the goroutine. See https://github.com/knative/eventing-sources/pull/124/commits/5592a9e34537d58bb304d99a6e5215f3d9381e54

pkg/apis/sources/v1alpha1/aws_sqs_types.go
srvaroa: (2) done https://github.com/knative/eventing-sources/pull/124/commits/a69666da0c2073fd2b4a17dafd07a7f23264bc0f

srvaroa: I'll resolve this one since there seems to be no objection to adding ARN support later.

pkg/adapter/awssqs/adapter.go
```diff
@@ -0,0 +1,211 @@
...
+
+// Adapter implements the AWS SQS adapter to deliver SQS messages from
+// an SQS queue to a Sink.
+type Adapter struct {
+
+	// Region is the AWS region
+	Region string
+
+	// QueueUrl is the AWS SQS URL that we're polling messages from
+	QueueUrl string
```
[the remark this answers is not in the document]
srvaroa: https://github.com/knative/eventing-sources/pull/124/commits/64fbe7337353449f8820c8b92fcc070f91c3ef95

pkg/adapter/awssqs/adapter.go
```diff
@@ -63,9 +60,17 @@ func (a *Adapter) Start(ctx context.Context, stopCh <-chan struct{}) error {
 
 	logger.Info("Starting with config: ", zap.Any("adapter", a))
 
+	urlParts := strings.Split(a.QueueURL, ".")
```
Harwayne: I would move this into a function `getRegion(a.QueueURL)` and provide an example URL to show what is expected (e.g. `https://sqs.us-east-2.amazonaws.com/123456789012/MyQueue`).

pkg/adapter/awssqs/adapter.go
```diff
@@ -82,43 +87,36 @@ func (a *Adapter) pollLoop(ctx context.Context, q *sqs.SQS, stopCh <-chan struct
 	logger := logging.FromContext(ctx)
 	cctx, cancel := context.WithCancel(ctx)
```
Harwayne: Not needed anymore.

pkg/adapter/awssqs/adapter.go
```diff
@@ -82,43 +87,36 @@ func (a *Adapter) pollLoop(ctx context.Context, q *sqs.SQS, stopCh <-chan struct
...
-						// get redelivered later, given that SQS is
-						// at-least-once delivery. That should be
-						// acceptable as "normal operation"
-						logger.Error("Failed to delete message", zap.Error(err))
-					}
+Loop:
+	for {
+		select {
+		case <-stopCh:
+			break Loop
```
Harwayne: What about returning here?

```golang
case <-stopCh:
    logger.Info("Exiting")
    return nil
```

pkg/adapter/awssqs/adapter_test.go
```diff
@@ -52,7 +51,7 @@ func TestPostMessage_ServeHTTP(t *testing.T) {
 			defer sinkServer.Close()
 
 			a := &Adapter{
-				QueueUrl:             "https://test.sqs.aws/123123",
+				QueueURL:             "https://test.sqs.aws/123123",
```
Harwayne: Because we now extract the region from the URL, let's make this more like a real URL, e.g. `https://sqs.us-east-2.amazonaws.com/123456789012/MyQueue`.

pkg/adapter/awssqs/adapter.go
srvaroa: Good point, I should have reviewed this after that change, thanks.

pkg/adapter/awssqs/adapter_test.go
srvaroa: https://github.com/knative/eventing-sources/pull/124/commits/573ca8b6814d24cd42ef3c01f50a7ec7bd76106b

pkg/adapter/awssqs/adapter.go
srvaroa: https://github.com/knative/eventing-sources/pull/124/commits/573ca8b6814d24cd42ef3c01f50a7ec7bd76106b

pkg/adapter/awssqs/adapter.go
srvaroa: https://github.com/knative/eventing-sources/pull/124/commits/573ca8b6814d24cd42ef3c01f50a7ec7bd76106b

pkg/adapter/awssqs/adapter.go
srvaroa: https://github.com/knative/eventing-sources/pull/124/commits/573ca8b6814d24cd42ef3c01f50a7ec7bd76106b

srvaroa:
@Harwayne there was a small change to the type tests required after the bump to /pkg which I forgot to add to the type. The change is simply applying the same fix to the newly added type and this should make tests pass now (needs the lgtm though)

Harwayne:
/lgtm 

And thanks for not rebasing! It made it much easier to see what changed 😄

srvaroa:
@Harwayne sorry, looks like it also needed an update-manifests (tests were passing locally so I didn't spot this)

grantr:
and sorry for the approval delay :(

srvaroa:
Thanks @grantr, no worries! Let me know if there's anything else I should fix to enable merging (cc @harwayne)

whynowy:
@srvaroa - one last request, please continue on the KIAM support :)

PR #127, opened 2018-11-21
Adding heartbeats source and receiver as a release-able component.

n3wscott:
## Proposed Changes

  * Release heartbeats as a source example with compiled image.
  * Release heartbeats-receiver as a compiled image.

**Release Note**

```release-note
Releasing heartbeats source and receiver. 
```

### DIFF
--- a/config/tools/heartbeats-receiver.yaml
+++ b/config/tools/heartbeats-receiver.yaml
@@ -1,0 +1,13 @@
+# Heartbeats receiver outputs the contents of a heartbeats cloud event as a
+# single log line.
+apiVersion: serving.knative.dev/v1alpha1
+kind: Service
+metadata:
+  name: heartbeats-receiver
+spec:
+  runLatest:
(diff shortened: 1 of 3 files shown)


evankanderson:
/assign @vaikas-google 

Can you coordinate with Ville, who I think was working on adding a Google Cloud Scheduler source?

At the very least, it seems like it might be useful to align the terminology -- I'd probably prefer to call this "periodic" or "cron" source rather than "heartbeat", which suggests a specific use-case.

n3wscott:
@vaikas-google where did we land on terminology?

vaikas-google:
well, so, there's Google Cloud Scheduler (vaikas-google/csr). There's also a cron in this very repository, so I think it's a bit confusing to have both, seems like the cron does superset of what this does. If we want to demonstrate the container source and keep it in the repo, perhaps it could be expanded to send a custom event type?

PR #129, opened 2018-11-23
fix: propagate GH headers into cloud event

### LINKED ISSUE #120: Github Source drops critical header
/kind bug

I am trying to use the `github.com/google/go-github` library to handle events received through the Github event source, and it seems that a key header is being dropped.

From my own logging:
```
ERROR: unable to parse webhook: unknown X-Github-Event in message:
```

What I'm doing is largely following the sample usage [here](https://godoc.org/github.com/google/go-github/github#ParseWebHook), with the exception that I don't call `ValidatePayload` because that secret verification is managed by the github event source.

Here's the code printing the above error:
```go
        event, err := github.ParseWebHook(github.WebHookType(r), payload)
        if err != nil {
                log.Printf("ERROR: unable to parse webhook: %v", err)
                http.Error(w, err.Error(), http.StatusBadRequest)
                return
        }
```

Making sure this sample usage works feels like it would be a great e2e test for this event source.

Note: I'm not sure if the "github source" or the "in-memory-channel" it proxies through is dropping the header.

srvaroa:
Fixes #120 

## Proposed Changes

Encode X-GitHub-Event and X-GitHub-Delivery in Cloud Event extensions, encoded according to:
- https://github.com/cloudevents/spec/blob/v0.1/spec.md#extensions
- https://github.com/cloudevents/spec/blob/v0.1/http-transport-binding.md#3131-http-header-names

### DIFF
--- a/pkg/githubsource/receive_adapter.go
+++ b/pkg/githubsource/receive_adapter.go
@@ -31,9 +31,8 @@ import (
 )
 
 const (
-	GHHeaderEvent     = "X-GitHub-Event"
-	GHHeaderDelivery  = "X-GitHub-Delivery"
-	CEExtensionPrefix = "CE-"
+	GHHeaderEvent    = "GitHub-Event"
+	GHHeaderDelivery = "GitHub-Delivery"
 )
 
 // GitHubReceiveAdapter converts incoming GitHub webhook events to
@@ -54,11 +53,11 @@ func (ra *GitHubReceiveAdapter) HandleEvent(payload interface{}, header webhooks
 
 func (ra *GitHubReceiveAdapter) handleEvent(payload interface{}, hdr http.Header) error {
 
-	gitHubEventType := hdr.Get(GHHeaderEvent)
-	eventID := hdr.Get(GHHeaderDelivery)
+	gitHubEventType := hdr.Get("X-" + GHHeaderEvent)
+	eventID := hdr.Get("X-" + GHHeaderDelivery)
 	extensions := map[string]interface{}{
-		CEExtensionPrefix + GHHeaderEvent:    hdr.Get(GHHeaderEvent),
-		CEExtensionPrefix + GHHeaderDelivery: hdr.Get(GHHeaderDelivery),
+		cloudevents.HeaderExtensionsPrefix + GHHeaderEvent:    hdr.Get("X-" + GHHeaderEvent),
+		cloudevents.HeaderExtensionsPrefix + GHHeaderDelivery: hdr.Get("X-" + GHHeaderDelivery),
 	}
 
 	log.Printf("Handling %s", gitHubEventType)
--- a/pkg/githubsource/receive_adapter.go
+++ b/pkg/githubsource/receive_adapter.go
@@ -30,6 +30,12 @@ import (
 	gh "gopkg.in/go-playground/webhooks.v3/github"
 )
 
+const (
+	GhHeaderEvent     = "X-GitHub-Event"
+	GhHeaderDelivery  = "X-GitHub-Delivery"
+	CeExtensionPrefix = "CE-"
+)
+
 // GitHubReceiveAdapter converts incoming GitHub webhook events to
 // CloudEvents and then sends them to the specified Sink
 type GitHubReceiveAdapter struct {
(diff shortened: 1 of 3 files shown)


pkg/githubsource/receive_adapter.go
```diff
@@ -30,6 +30,12 @@ import (
 	gh "gopkg.in/go-playground/webhooks.v3/github"
 )
 
+const (
+	GhHeaderEvent     = "X-GitHub-Event"
+	GhHeaderDelivery  = "X-GitHub-Delivery"
+	CeExtensionPrefix = "CE-"
```
mattmoor: nit: Go style wants acronyms to be all caps e.g. `CEExtensionPrefix`.

mattmoor: `s/all caps/consistent caps/`, if they lead a private variable then all lowercase is fine too.

pkg/githubsource/receive_adapter.go
```diff
@@ -61,16 +68,20 @@ func (ra *GitHubReceiveAdapter) handleEvent(payload interface{}, gitHubEventType
...
 
-func (ra *GitHubReceiveAdapter) postMessage(payload interface{}, source, eventType, eventID string) error {
+func (ra *GitHubReceiveAdapter) postMessage(payload interface{}, source, eventType, eventID string, hdr http.Header) error {
 	ctx := cloudevents.EventContext{
 		CloudEventsVersion: cloudevents.CloudEventsVersion,
 		EventType:          eventType,
 		EventID:            eventID,
 		EventTime:          time.Now(),
 		Source:             source,
+		Extensions: map[string]interface{}{
```
mattmoor: I'm guessing that this is the key bit?  Thanks for the fix!

srvaroa: Yeah, my understanding of the cloud event specs is that Extensions is the right place to propagate this kind of metadata. Although, note that they mandate CE-X as the prefix for the keys, this means that the consumer downstream will have to s/CE-// to get back the headers wanted by the GH sdk.

pkg/githubsource/receive_adapter.go
```diff
@@ -30,6 +30,12 @@ import (
 	gh "gopkg.in/go-playground/webhooks.v3/github"
 )
 
+const (
+	GHHeaderEvent     = "X-GitHub-Event"
+	GHHeaderDelivery  = "X-GitHub-Delivery"
+	CEExtensionPrefix = "CE-"
```
matzew: how about reusing this ? 
https://github.com/knative/pkg/blob/master/cloudevents/encoding_binary.go#L60

pkg/githubsource/receive_adapter.go
```diff
@@ -61,16 +68,20 @@ func (ra *GitHubReceiveAdapter) handleEvent(payload interface{}, gitHubEventType
 	cloudEventType := fmt.Sprintf("%s.%s", sourcesv1alpha1.GitHubSourceEventPrefix, gitHubEventType)
 	source := sourceFromGitHubEvent(gh.Event(gitHubEventType), payload)
 
-	return ra.postMessage(payload, source, cloudEventType, eventID)
+	return ra.postMessage(payload, source, cloudEventType, eventID, hdr)
 }
 
-func (ra *GitHubReceiveAdapter) postMessage(payload interface{}, source, eventType, eventID string) error {
+func (ra *GitHubReceiveAdapter) postMessage(payload interface{}, source, eventType, eventID string, hdr http.Header) error {
```
matzew: isn't `eventId` in the headers ?

pkg/githubsource/receive_adapter.go
srvaroa: I didn't use it as it's private, but if you think it makes more sense I can push a PR over there.

srvaroa:
@mattmoor PR sent https://github.com/knative/pkg/pull/172, I'll update here if it gets merged.

pkg/githubsource/receive_adapter.go
srvaroa: I made a mess in these methods, fixing it.

mattmoor:
@matzew @srvaroa I just approved the pkg PR

pkg/githubsource/receive_adapter.go
srvaroa: Expanding on this: I think it still makes sense to propagate both GH custom headers even if X-GitHub-Delivery is also put in the cloud event id. The GH sdk will typically expect the original headers to be there. If we remove the -Delivery one, a consumer downstream would have to reinject it based on the cloud event ID, which not obvious. On the other hand, recovering all CE-X- headers blindly would immediately be SDK friendly. Does that make sense?

srvaroa:
Updated with the latest pkg (which includes the now public constant)

srvaroa:
@mattmoor the failures after upgrading pkg seem to relate to https://github.com/knative/pkg/commit/4b704fa7948ad9ae8ec90d1cd5b4a34516b252ea, do you have any pointers on the right way to fix them? Thx in advance.

srvaroa:
I found a commit that seems to deal with this same update (https://github.com/knative/eventing/commit/7dc7624e34a4da510648d5d58235638502dd06a9). I've pushed the bump on knative/pkg separately at https://github.com/knative/eventing-sources/pull/130, applying the same fix to the tests to let them run. I'll wait until that is merged and update this PR.

pkg/githubsource/receive_adapter.go
```diff
@@ -40,16 +45,21 @@ type GitHubReceiveAdapter struct {
...
 }
 
-func (ra *GitHubReceiveAdapter) handleEvent(payload interface{}, gitHubEventType string, eventID string) error {
+func (ra *GitHubReceiveAdapter) handleEvent(payload interface{}, hdr http.Header) error {
+
+	gitHubEventType := hdr.Get("X-" + GHHeaderEvent)
+	eventID := hdr.Get("X-" + GHHeaderDelivery)
+	extensions := map[string]interface{}{
+		cloudevents.HeaderExtensionsPrefix + GHHeaderEvent:    hdr.Get("X-" + GHHeaderEvent),
+		cloudevents.HeaderExtensionsPrefix + GHHeaderDelivery: hdr.Get("X-" + GHHeaderDelivery),
```
matzew: nice!

srvaroa:
@mattmoor this one was just pending the bump on pkg which seems to be working fine, any changes you'd like to see before merging?

mattmoor:
I am now super excited for 0.3!

PR #137, opened 2018-12-03
Use CR unstructured client for dynamic objects

grantr:
Since https://github.com/kubernetes-sigs/controller-runtime/pull/101, controller-runtime will create and manage dynamic clients when given `unstructured.Unstructured` objects with a filled in `TypeMeta`. Using this allows us to eliminate the dynamic clients created to deal with Addressable objects.

Also updates tests to call the interface Addressable instead of Sinkable.

### DIFF
--- a/pkg/controller/containersource/reconcile.go
+++ b/pkg/controller/containersource/reconcile.go
@@ -37,17 +37,15 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 type reconciler struct {
-	client        client.Client
-	scheme        *runtime.Scheme
-	dynamicClient dynamic.Interface
-	recorder      record.EventRecorder
+	client   client.Client
+	scheme   *runtime.Scheme
+	recorder record.EventRecorder
 }
 
 // Reconcile compares the actual state with the desired, and attempts to
@@ -84,7 +82,7 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 		ServiceAccountName: source.Spec.ServiceAccountName,
 	}
 
-	err = r.setSinkURIArg(source, args)
+	err = r.setSinkURIArg(ctx, source, args)
 	if err != nil {
 		return source, err
 	}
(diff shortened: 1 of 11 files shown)


pkg/controller/sdk/provider.go
```diff
@@ -30,6 +31,7 @@ import (
 type KnativeReconciler interface {
 	Reconcile(ctx context.Context, object runtime.Object) (runtime.Object, error)
 	InjectClient(c client.Client) error
+	//TODO(grantr) make these separate interfaces
 	InjectConfig(c *rest.Config) error
```
Harwayne: Does anything need this anymore? I thought it was only used for the dynamic client.

grantr: I had originally pulled this out into a separate PR, but it was complicated to make this change separately. Further commits forthcoming.

grantr: /hold cancel

PTAL.

pkg/controller/sdk/provider.go
```diff
@@ -29,8 +28,6 @@ import (
 
 type KnativeReconciler interface {
 	Reconcile(ctx context.Context, object runtime.Object) (runtime.Object, error)
-	InjectClient(c client.Client) error
```
Harwayne: I'm surprised this line can be removed. Are any of these reconcilers viable without a `client`?

grantr: The `inject.ClientInto` method will detect if the reconciler implements `InjectClient` and call it if so.

There's no need to require this method in the interface because there are other ways of injecting the client. For example, in `pkg/controller/containersource/provider.go`:

```diff
   p := &sdk.Provider{
     AgentName: controllerAgentName,
     Parent:    &v1alpha1.ContainerSource{},
     Owns:      []runtime.Object{&appsv1.Deployment{}},
     Reconciler: &reconciler{
       recorder: mgr.GetRecorder(controllerAgentName),
       scheme:   mgr.GetScheme(),
+      client:  mgr.GetClient(),
     },
   }
```

Harwayne: So my preference is to continue to require it in the interface. You are right that there are other ways to do it, but do they provide any advantages?

If an author forgets to somehow inject the client (via whichever method they prefer), then this controller will fail at runtime. And even then, only after the controller is trying to do something. I'm a fan of keeping this a compile-time error 😄

pkg/controller/containersource/reconcile.go
```diff
@@ -216,7 +214,5 @@ func (r *reconciler) InjectClient(c client.Client) error {
 }
 
 func (r *reconciler) InjectConfig(c *rest.Config) error {
```
Harwayne: Can this be removed?

grantr: Done.

PR #144, opened 2018-12-05
Allow deleting a GitHubSource with a missing sink

grantr:
Fixes https://github.com/knative/eventing/issues/672.

There's probably a larger refactoring needed to make the logic cleaner, but this smaller fix should get in more quickly.

## Proposed Changes

  * Don't return early if sink is missing and source is deleted

**Release Note**

```release-note
GitHubSource objects can now be deleted even if their sink is missing.
```

### DIFF
--- a/pkg/controller/githubsource/reconcile.go
+++ b/pkg/controller/githubsource/reconcile.go
@@ -95,9 +95,12 @@ func (r *reconciler) reconcile(ctx context.Context, source *sourcesv1alpha1.GitH
 	uri, err := sinks.GetSinkURI(ctx, r.client, source.Spec.Sink, source.Namespace)
 	if err != nil {
 		source.Status.MarkNoSink("NotFound", "%s", err)
-		return err
+		if !deleted {
+			return err
+		}
+	} else {
+		source.Status.MarkSink(uri)
 	}
-	source.Status.MarkSink(uri)
 
 	ksvc, err := r.getOwnedService(ctx, source)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/githubsource/reconcile.go
```diff
@@ -95,9 +95,12 @@ func (r *reconciler) reconcile(ctx context.Context, source *sourcesv1alpha1.GitH
 	uri, err := sinks.GetSinkURI(ctx, r.client, source.Spec.Sink, source.Namespace)
 	if err != nil {
```
vaikas-google: Should we check the error for NotFound and not bail on any error?

neosab:
@grantr I understand the targeted fix for sinks but should we also do the same for secrets because most likely someone would be doing `kubectl delete -f sources_v1alpha1_githubsource.yaml` and probably this time the secret is deleted ahead of the channel.

grantr:
Good point @neosab. A deleted secret is trickier because there's no way to delete the webhook from GitHub if the secret is gone. At that point it's probably best to give up and remove the finalizer while creating a corev1.Event that warns of a possible orphaned webhook in GitHub.

pkg/controller/githubsource/reconcile.go
neosab: `GetSinkURI` also returns an error when sink ref is nil which can be ignored in this deleted case. Although I doubt this scenario could ever happen when the finalizer is set (since we would have previously validated sink ref existed while adding the finalizer)

grantr:
Replacing this with #146. That PR increases the scope to handle a missing secret as well, and refactors things to be more clear about finalizer failure handling.

sebgoa:
Can you delete the source if the secret is missing ? Not deleted after source creation, but missing all together in the first place or even if the keys are not found .

grantr:
@sebgoa with #146 the source can be deleted even if secrets are missing, but the webhook (if created) will be orphaned.

PR #146, opened 2018-12-06
Split GitHubSource finalizer into separate method

grantr:
The finalization pass only cares about the access token and deleting the webhook. Splitting it into a separate method simplifies both the reconcile and finalize flows.

Now the controller will attempt to delete the webhook only once. If the first attempt fails, the finalizer is removed anyway. In this case an event is recorded with the webhook id so the operator can delete the webhook manually if desired.

Removal of the finalizer no longer depends on the addressable being present.

Fixes https://github.com/knative/eventing/issues/672.
Replaces #144.

**Release Note**
```release-note
GitHubSource objects can now be deleted if their sink or secret are missing.
In case of webhook deletion failure, a Kubernetes event will be generated
containing the webhook's id so it can be deleted manually.
```

### DIFF
--- a/pkg/controller/githubsource/reconcile.go
+++ b/pkg/controller/githubsource/reconcile.go
@@ -70,14 +70,18 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 		logger.Warnf("Failed to get metadata accessor: %s", zap.Error(err))
 		return object, err
 	}
-	deleted := accessor.GetDeletionTimestamp() != nil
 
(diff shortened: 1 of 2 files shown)


grantr:
Also changed in this PR: adding the finalizer is delayed until the webhook is ready to be created, that is, the secrets have been found, the sink URI has been identified, and the service is ready. If any of those steps fail terminally, no finalizer is added and the source can be deleted normally.

bbrowning:
Thanks for chasing down these loose ends!

PR #150, opened 2018-12-07
Format markdown

mattmoor-sockpuppet:
Produced via: `prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github)`


mattmoor:
/ok-to-test
/assign @evankanderson @grantr 

Eventing Sources is a new addition, so this is the bigger initial one.

PR #162, opened 2019-01-02
Improve the error message when sink is unable to find an address

evankanderson:
## Proposed Changes

  * Include "namespace/name (apiVersion.Kind)" in the error object when an Addressable can't be handled.

### DIFF
--- a/pkg/controller/sinks/sinks.go
+++ b/pkg/controller/sinks/sinks.go
@@ -41,18 +41,20 @@ func GetSinkURI(ctx context.Context, c client.Client, sink *corev1.ObjectReferen
 		return "", err
 	}
 
+	objIdentifier := fmt.Sprintf("\"%s/%s\" (%s.%s)", u.GetNamespace(), u.GetName(), u.GetAPIVersion(), u.GetKind())
+
 	t := duckv1alpha1.AddressableType{}
 	err = duck.FromUnstructured(u, &t)
 	if err != nil {
-		return "", fmt.Errorf("failed to deserialize sink: %v", err)
+		return "", fmt.Errorf("failed to deserialize sink %s: %v", objIdentifier, err)
 	}
 
 	if t.Status.Address == nil {
-		return "", fmt.Errorf("sink does not contain address")
+		return "", fmt.Errorf("sink %s does not contain address", objIdentifier)
 	}
 
 	if t.Status.Address.Hostname == "" {
-		return "", fmt.Errorf("sink contains an empty hostname")
+		return "", fmt.Errorf("sink %s contains an empty hostname", objIdentifier)
 	}
 
 	return fmt.Sprintf("http://%s/", t.Status.Address.Hostname), nil


pkg/controller/containersource/reconcile_test.go
```diff
@@ -94,7 +94,7 @@ var testCases = []controllertesting.TestCase{
 		},
 		ReconcileKey: fmt.Sprintf("%s/%s", testNS, containerSourceName),
 		Scheme:       scheme.Scheme,
-		WantErrMsg:   "sink does not contain address",
+		WantErrMsg:   "sink \"testnamespace/testunaddressable\" (duck.knative.dev/v1alpha1.KResource) does not contain address",
```
grantr: I suggest changing the surrounding quotes to backticks and unescaping the inner quotes. Mostly for consistency with line 148, but also the backticks read cleaner to me.

Optionally, remove the inner quotes entirely since spaces aren't allowed in object names.

pkg/controller/sinks/sinks.go
```diff
@@ -41,18 +41,20 @@ func GetSinkURI(ctx context.Context, c client.Client, sink *corev1.ObjectReferen
 		return "", err
 	}
 
+	objIdentifier := fmt.Sprintf("\"%s/%s\" (%s.%s)", u.GetNamespace(), u.GetName(), u.GetAPIVersion(), u.GetKind())
```
grantr: Optional: You can use backticks here too to avoid having to escape inner quotes, if that's your bag.

pkg/controller/githubsource/reconcile_test.go
```diff
@@ -98,7 +98,7 @@ var testCases = []controllertesting.TestCase{
 		},
 		ReconcileKey: fmt.Sprintf("%s/%s", testNS, gitHubSourceName),
 		Scheme:       scheme.Scheme,
-		WantErrMsg:   "sink does not contain address",
+		WantErrMsg:   "sink \"testnamespace/testunaddressable\" (duck.knative.dev/v1alpha1.KResource) does not contain address",
```
grantr: Suggest backticks instead.

pkg/controller/githubsource/reconcile_test.go
```diff
@@ -133,13 +133,13 @@ var testCases = []controllertesting.TestCase{
 			func() runtime.Object {
 				s := getGitHubSource()
 				s.Status.InitializeConditions()
-				s.Status.MarkNoSink("NotFound", "sink does not contain address")
+				s.Status.MarkNoSink("NotFound", "sink \"testnamespace/testsink\" (duck.knative.dev/v1alpha1.Sink) does not contain address")
```
grantr: Suggest backticks instead.

pkg/controller/sinks/sinks.go
evankanderson: I'm happy to use backticks in tests, where I do think it makes things more readable, but I'm inclined to use double-quotes and escaping here so that readers don't need to remember if there's anything special about <code>`</code>.

vaikas-google: instead of manually creating the u.GetAPIVersion, u.GetKind, it might be better to just use:
u.GroupVersionKind()?
Or, it's really no different from:
sink.GeroupVersionKind so it's got the standard k8s formatted string.

evankanderson: Hey, I hadn't seen `GroupVersionKind()`. Switched, but it makes the output slightly more verbose.

PR #163, opened 2019-01-02
Move GCPPubSub source to "contrib" in preparation from removing it from core.

evankanderson:
## Proposed Changes

  * Move the gcppubsub controller to a separate statefulset. 
  * Create separate RBAC rules for the gcppubsub controller.

We should do the same with the AWS source (at least).

**Release Note**

```release-note
The GCPPubSub source has been moved to a separate controller and set of release YAML files.
```


grantr:
I'll let @Harwayne lgtm this, but minor note: a change to https://github.com/knative/eventing-sources/blob/master/hack/release.sh#L25 might also be required.

contrib/gcppubsub/config/manager.yaml
```diff
@@ -0,0 +1,57 @@
...
+  labels:
+    control-plane: gcppubsub-controller
+spec:
+  selector:
+    control-plane: gcppubsub-controller
+  ports:
+  - port: 443
+---
+apiVersion: apps/v1
+kind: StatefulSet
```
Harwayne: Can be a standard `Deployment`.

contrib/gcppubsub/config/rbac_role_binding.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  creationTimestamp: null
+  name: gcppubsub-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: gcppubsub-controller-role
+subjects:
+- kind: ServiceAccount
+  name: default
```
Harwayne: Should we also create a unique service account?

Harwayne:
/lgtm
/hold
Holding for the two comments, which are optional, just want to make sure you see them first.

Note that this will also miss some of the mark down files that point at specific paths. E.g. `contrib/gcppubsub/samples/README.md` in which every `ko` and `kubectl` path is incorrect.

contrib/gcppubsub/config/manager.yaml
evankanderson: If we end up scaling horizontally beyond 1 replica, I think the controllers would be stateful in that each one would cover a different range of the keyspace.

contrib/gcppubsub/config/rbac_role_binding.yaml
evankanderson: Good point, done.

I also added the necessary startup secret (as an empty string) to make initial install start without waiting for the secret to exist.

contrib/gcppubsub/samples/README.md
```diff
@@ -12,112 +20,110 @@
...
-     it.
-1. Create a `Channel`. You can use your own `Channel` or use the provided
-   sample, which creates `qux-1`. If you use your own `Channel` with a different
-   name, then you will need to alter other commands later.
+   [GCP PubSub Source from this yaml](../config/default-gcppubsub.yaml) from
+   source:
 
    ```shell
-   kubectl -n default apply -f samples/gcp-pubsub-source/channel.yaml
+   ko apply --filename https://github.com/knative/eventing-sources/tree/master/contrib/gcppubsub/config/default-gcppubsub.yaml
```
Harwayne: My preference is to use local paths (relative to repo root), so that these instructions can be used during development. Then the docs repo would use the released versions.

And what does it mean to `ko apply` the yaml from master? Does that mean any local changes to the binary would be honored, but not from the yaml?

contrib/gcppubsub/samples/README.md
```diff
@@ -12,112 +20,110 @@
...
+       # Knative Eventing. The following command will overwrite it. If you don't
+       # want to overwrite it, then skip this command.
+       kubectl -n knative-sources create secret generic gcppubsub-source-key --from-file=key.json=knative-source.json --dry-run -o yaml | kubectl apply --filename -
+
+       # The second secret should not already exist, so just try to create it.
+       kubectl -n default create secret generic google-cloud-key --from-file=key.json=knative-source.json
+       ```
+
+       `gcppubsub-source-key` and `key.json` are pre-configured values in the
+       `controller-manager` StatefulSet which manages your Eventing sources.
```
Harwayne: gcppubsub-manager

contrib/gcppubsub/samples/gcp-pubsub-source.yaml
```diff
@@ -13,6 +13,6 @@ spec:
   googleCloudProject: MY_GCP_PROJECT
   topic: TOPIC_NAME
   sink:
-    apiVersion: eventing.knative.dev/v1alpha1
-    kind: Channel
-    name: qux-1
+    apiVersion: serving.knative.dev/v1alpha1
```
Harwayne: At this point should we just use a `Deployment` and K8s `Service` instead?

contrib/gcppubsub/samples/README.md
```diff
@@ -12,112 +20,110 @@
...
-   controller.
-
-   ```shell
-   ko -n default apply -f config/default-gcppubsub.yaml
-   ```
-
-   - Note that if the `Source` Service Account secret is in a non-default
-     location, you will need to update the YAML first.
-
 1. Replace the place holders in `gcp-pubsub-source.yaml`.
```
Harwayne: I recommend doing this as setting environment variables, then using sed on the yaml. So that people don't have to manipulate the file directly.

```shell
export TOPIC_NAME=laconia

sed -e "s/MY_GCP_PROJECT/$PROJECT_ID/g" -e "s/TOPIC_NAME/$TOPIC_NAME/g" gcp-pubsub-source.yaml | kubectl apply -
```

contrib/gcppubsub/samples/README.md
evankanderson: Maybe this sample should go in favor of the public docs? (I ended up copying about 1/3rd of the public docs in here, because they have a much clearer flow and more repeatable command-line flow for getting the GCP side set up.)

evankanderson: It's not clear whether these docs are development instructions (at which point, they are probably too detailed) or usage instructions that we'd point end-users at. Right now they are awkwardly between the two.

contrib/gcppubsub/samples/README.md
evankanderson: `gcppubsub-controller`. 😁

contrib/gcppubsub/samples/README.md
evankanderson: Done, thanks!

evankanderson:
Updated, PTAL.

contrib/gcppubsub/samples/README.md
Harwayne: Export PROJECT_ID and TOPIC_NAME.

contrib/gcppubsub/samples/README.md
evankanderson: I think we will move the instructions from knative/docs if and when we move this source to the GoogleCloudPlatform org. At that point, these instructions will merge with the knative/docs version.

I almost deleted this sample, but I thought that it might be worth exploring how small the documentation could be in terms of objects, and I think this does simplify the footprint in terms of the documentation. I removed the link to Knative serving here now, too.

contrib/gcppubsub/samples/README.md
evankanderson: Done -- thanks for prodding me to find a copy-pasteable way to get the project ID.

PR #171, opened 2019-01-08
fix compilation of gcppubsub adapter test

dprotaso:


Harwayne:
/lgtm

Looks like it was broken by #166. How did prow let it in? Does prow not check things in `contrib/`?

dprotaso:
> /lgtm
> 
> Looks like it was broken by #166. How did prow let it in? Does prow not check things in `contrib/`?

Yeah looks like only things instead `pkg` and `cmd` are unit tested
https://github.com/knative/eventing-sources/blob/master/test/presubmit-tests.sh#L27

grantr:
Does this need to be in the 0.3 release?

grantr:
Answer: no because it's tests only.

PR #172, opened 2019-01-08
[WIP] Enable the use of the status subresource

dprotaso:
Few odd things

The following weren't updated via scripts and I did it manually
- `config/default-awssqs.yaml`
- `contrib/gcppubsub/config/rbac_role.yaml`


grantr:
Here's the "documentation" for the `--input-dir` option I mentioned: https://github.com/kubernetes-sigs/controller-tools/blob/master/cmd/controller-gen/main.go

evankanderson:
It turns out that I missed the `controller-gen` update in update-manifests. I have a PR for that separately.

evankanderson:
Thanks for waiting... see if your set of changes is smaller now!

pkg/controller/kuberneteseventsource/doc.go
```diff
@@ -16,7 +16,9 @@ limitations under the License.
 
 // Package kuberneteseventsource implements a KubernetesEventSource controller.
 // +kubebuilder:rbac:groups=sources.eventing.knative.dev,resources=kuberneteseventsources,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=sources.eventing.knative.dev,resources=kuberneteseventsources/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=sources.eventing.knative.dev,resources=containersources,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=sources.eventing.knative.dev,resources=containersources/status,verbs=get;update;patch
```
evankanderson: I don't think we need to `updateStatus` on ContainerSources in the KubernetesEventSource -- the k8s EventSource is creating the ContainerSource.

PR #173, opened 2019-01-08
Update update-manifests to look harder in contrib.

evankanderson:
Fixes bugs noticed in #172 

## Proposed Changes

  * Run controller-gen in the `contrib/gcppubsub` directory, and use the results rather than hand-hacking.
  * Actually run `kustomize` for `awssqs`
  * Remove the vestigial `contrib/default-gcppubsub.yaml` file.


evankanderson:
/hold cancel
/assign @Harwayne 

I think this is ready to go now.

grantr:
Does this need to be in the 0.3 release?

evankanderson:
No, I think it's fine to leave this for 0.4 -- it changes the name of one ClusterRoleBinding, and removes a few permissions from the `default-awssqs` controller installation; the extra permissions should be harmless, and the rename of the ClusterRoleBinding is cosmetic, since neither should collide with anything.

contrib/gcppubsub/config/rbac_role.yaml
```diff
@@ -29,25 +29,24 @@ rules:
...
+  - channels
   verbs:
   - get
   - list
   - watch
-- apiGroups:
-  - ""
-  resources:
-  - secrets
+- resources:
```
Harwayne: This doesn't have an apiGroups, is that on purpose?

Caused by doc.go bug.

contrib/gcppubsub/pkg/controller/doc.go
```diff
@@ -17,5 +17,11 @@ limitations under the License.
 // Package gcppubsub implements the GcpPubSubSource controller.
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=sources.eventing.knative.dev,resources=gcppubsubsources,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=,resources=secrets,verbs=get;list;watch
+
+// This is needed to be able to watch channels when delivering to a channel.
 // +kubebuilder:rbac:groups=eventing.knative.dev,resources=channels,verbs=get;list;watch
+
+// This is needed to be able to watch the sink when delivering messages directly.
+// +kubebuilder:rbac:gcoups=serving.knative.dev,resources=services;routes,verbs=get;list;watch
```
Harwayne: gcoups -> groups

contrib/gcppubsub/config/rbac_role_binding.yaml
```diff
@@ -2,12 +2,12 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   creationTimestamp: null
-  name: gcppubsub-controller
+  name: gcppubsub-controller-rolebinding
```
Harwayne: FWIW I find the `-rolebinding` suffix redundant.

evankanderson: You'd need to update the vendor:

https://github.com/knative/eventing-sources/blob/master/vendor/sigs.k8s.io/controller-tools/pkg/generate/rbac/manifests.go#L52

contrib/gcppubsub/pkg/controller/doc.go
evankanderson: Thanks, that fixed the other issue.

evankanderson:
Sorry, I'd been trying to remove the edit/un-edit on the files and got into a place where I had to `--amend` and `push --force`. I'll try to avoid doing that in the future.

dprotaso:
@Harwayne @evankanderson 

If you click on the `forced-pushed` link you'll see the smaller diff you're looking for
ie. https://github.com/knative/eventing-sources/compare/efa07cf94ebe647951d62044c6aa9e471a6a30d8..95d1a41cf99b70d64753ba8a54500c6cce8a5a6e

PR #121, opened 2018-11-19
Add E2E test

grantr:
This is a copy of the eventing E2E test which checks if KubernetesEventSource can deliver an event to a sink.

This isn't the final form for E2E tests of sources, but a stopgap solution to prevent the eventing-sources nightly release silently breaking eventing E2E tests.

PR #178, opened 2019-01-11
Another attempt to add an E2E test

bbrowning:
This is an attempt to push #121 to completion so we get some basic E2E testing for eventing-sources. It may take a couple of iterations here to get things passing.


vaikas-google:
@n3wscott since you just cleaned up eventing e2e tests.

test/crd.go
```diff
@@ -0,0 +1,195 @@
...
+				{
+					ConfigurationName: configName,
+					Percent:           100,
+				},
+			},
+		},
+	}
+}
+
+// Configuration returns a Configuration object in namespace with the name names.Config
```
vaikas-google: nit. this comment is a bit confusing. names.Config doesn't seem to exist?

test/e2e/e2e.go
```diff
@@ -0,0 +1,234 @@
...
+			Kind:     "ClusterRole",
+			Name:     "cluster-admin",
+			APIGroup: "rbac.authorization.k8s.io",
+		},
+	}
+	err = CreateClusterRoleBinding(clients, crb, logger, cleaner)
+	if err != nil {
+		return err
+	}
+	return nil
```
vaikas-google: Do we not want these cleaned up? As in, how come these are not added to the cleaner?

test/e2e/e2e.go
```diff
@@ -0,0 +1,234 @@
...
+		logs, err := PodLogs(clients, podName, containerName, logger)
+		if err != nil {
+			return true, err
+		}
+		return strings.Contains(string(logs), content), nil
+	})
+}
+
+// WaitForAllPodsRunning will wait until all pods in the given namespace are running
+func WaitForAllPodsRunning(clients *test.Clients, logger *logging.BaseLogger, namespace string) error {
```
vaikas-google: Does this block forever or does it timeout at some point? Or put another way, if something goes wrong, will things block forevs?

test/e2e/k8s_events_test.go
```diff
@@ -0,0 +1,105 @@
...
+	err = CreatePod(clients, test.NGinxPod(testNamespace), logger, cleaner)
+	if err != nil {
+		t.Fatalf("Failed to create Pod: %v", err)
+	}
+
+	WaitForAllPodsRunning(clients, logger, testNamespace)
+
+	err = WaitForLogContent(clients, logger, routeName, "user-container", "Created container")
+	if err != nil {
+		t.Fatalf("Events for container created not received: %v", err)
```
vaikas-google: bit of a nit, my understanding has been that errors should be logged with %s.

n3wscott:
It looks like there is a lot of copy/pasta from the e2e test from eventing. I think we can let this in at the moment but I think we (knative) should build up a e2e framework that can be shared between the project. What do you think of that overall plan? 

So, let this in, but it is tech-debt.

test/e2e/k8s_events_test.go
n3wscott: we have always used `%v` I think, when did this change?

bbrowning:
There is definitely copy & paste here - both from the previous PR (121) and from Eventing repo in general. At the time, my goal was to just push the e2e work already started to a working state so that we'd get some e2e test coverage here.

I've resolved the merge conflicts here so it's unblocked from getting in. And I'll create an issue in knative/pkg to track pulling these common e2e test bits out. I believe other repos also have similar duplication of e2e helpers that would also benefit from consolidation.

bbrowning:
@vaikas-google @n3wscott Note that the most recent unit tests run as part of this PR exposed the same data race observed on #191 - a lot of `DATA RACE` mentions at https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_eventing-sources/178/pull-knative-eventing-sources-unit-tests/1093897152024809473/build-log.txt

n3wscott:
minor conflicts^

n3wscott:
which one wins? https://github.com/knative/eventing-sources/pull/121

test/cleanup.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright 2018 The Knative Authors
```
vaikas-google: nit, 2019 in all the new files.

test/e2e/k8s_events_test.go
bbrowning: It looks like `%v` is used all over the places in other knative repos when using `t.Fatalf`, so I'll leave this as-is for now.

test/e2e/e2e.go
bbrowning: They are already added to the cleaner from the `CreateServiceAccount` and `CreateClusterRoleBinding` methods called here.

test/e2e/e2e.go
bbrowning: This timeouts after 8 minutes - the `WaitForPodListState` method called here is in knative/pkg and the timeout is set there.

bbrowning:
This initial e2e test is passing again and ready for an updated review. The expectation is that once we get an e2e suite bootstrapped others can start adding e2e tests for each source.

This is a full E2E test of the Kubernetes Events source wired up via Eventing. I know we've moved towards reducing test dependencies between repos in the other projects, but I left that dependency in for now as otherwise there's nowhere that confirms Eventing Sources and Eventing actually work together.

I believe the existing issue https://github.com/knative/pkg/issues/26 is probably the best place to track reducing this kind of copy/paste between e2e setups in each repo and I added a comment there as well as this comment that will link the two issues.

test/clients.go
```diff
@@ -0,0 +1,85 @@
...
+
+	clients.Sources, err = sources.NewForConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return clients, nil
+}
+
+func buildClientConfig(kubeConfigPath string, clusterName string) (*rest.Config, error) {
```
srinivashegde86: BuildclientConfig() can be used from pkg/test

test/cleanup.go
```diff
@@ -0,0 +1,122 @@
...
+			}); err != nil {
+				c.logger.Errorf("Error: %v", err)
+			}
+		}
+	}
+	return nil
+}
+
+// CleanupOnInterrupt will execute the function cleanup if an interrupt signal is caught
+func CleanupOnInterrupt(cleanup func(), logger *logging.BaseLogger) {
```
srinivashegde86: CleanupOnInterrupt() can be used from pkg/test

test/e2e/e2e.go
```diff
@@ -0,0 +1,234 @@
+package e2e
```
srinivashegde86: missing file copyright header

test/e2e/e2e.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/knative/eventing/pkg/apis/eventing/v1alpha1"
+	pkgTest "github.com/knative/pkg/test"
+	"github.com/knative/pkg/test/logging"
+	servingV1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	rbacV1beta1 "k8s.io/api/rbac/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	// Mysteriously required to support GCP auth (required by k8s libs).
```
srinivashegde86: I dont think we need this anymore

bbrowning: Serving and Eventing master still have this in as well. I can attempt to take it out and see what happens, but I assume something still needs it.

bbrowning:
Thanks for the reviews!

PR #179, opened 2019-01-11
Soft-revert #137 "Use CR unstructured client for dynamic objects"

grantr:
This is an attempt to track down the issues caused by https://github.com/knative/eventing/issues/718.

It creates two versions of GetSinkURI, pre-#137 (dynamic clients) and post-#137 (controller-runtime unstructured client). The older sources have been reverted to use the dynamic clients version. The cronjobsource and awssqssource never had dynamic client versions, so they still use the controller-runtime unstructured client version.

If the eventing e2e tests pass with this change, then it seems clear that #137 caused them.


evankanderson:
Looks like this is not needed.

PR #183, opened 2019-01-16
add label indicating these are knative eventing sources

### LINKED ISSUE #176: make it easier to see installed Source CRs in a cluster.
## Expected Behavior
it's a bit tricky to see what CRs for sources have been installed. by convention if we just label the created CRs consistently, we'll be able to provide a cleaner user experience.

For example, I can see available sources because I have tagged the CR created like so:

https://github.com/vaikas-google/csr/blob/master/config/cloudschedulersource.yaml#L7

vaikas@penguin:~/projects/go/src/github.com/vaikas-google$ kubectl get crds -l "eventing.knative.dev/source=true"
NAME                                      AGE
cloudschedulersources.sources.aikas.org   50d
gcssources.sources.aikas.org              21h

## Actual Behavior
you have to iterate through all the CRDs to see if they are sources. For example, these all come from the eventing-sources but are not easily discoverable.

vaikas@penguin:~/projects/go/src/github.com/vaikas-google$ kubectl get crds | grep sources.eventing
containersources.sources.eventing.knative.dev         58d
cronjobsources.sources.eventing.knative.dev           7h
gcppubsubsources.sources.eventing.knative.dev         58d
githubsources.sources.eventing.knative.dev            58d
kuberneteseventsources.sources.eventing.knative.dev   58d

## Additional Info
I'm not wedded to the particular label, but sooner we agree on a convention, better off we are.

vaikas-google:
Fixes #176 

## Proposed Changes

  * Add label     eventing.knative.dev/source: "true"
  *
  *

**Release Note**

```release-note
Each source CRD is now labeled with:
    eventing.knative.dev/source: "true"
which means you can find list of available sources with:
 kubectl get crds -l "eventing.knative.dev/source=true"
```

### DIFF
--- a/config/crds/sources_v1alpha1_awssqssource.yaml
+++ b/config/crds/sources_v1alpha1_awssqssource.yaml
@@ -4,6 +4,7 @@ metadata:
   creationTimestamp: null
   labels:
     controller-tools.k8s.io: "1.0"
+    eventing.knative.dev/source: "true"
   name: awssqssources.sources.eventing.knative.dev
 spec:
   group: sources.eventing.knative.dev
--- a/config/crds/sources_v1alpha1_awssqssource.yaml
+++ b/config/crds/sources_v1alpha1_awssqssource.yaml
@@ -4,7 +4,6 @@ metadata:
   creationTimestamp: null
   labels:
     controller-tools.k8s.io: "1.0"
-    eventing.knative.dev/source: "true"
   name: awssqssources.sources.eventing.knative.dev
 spec:
   group: sources.eventing.knative.dev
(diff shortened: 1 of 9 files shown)


grantr:
Since these files are generated, I think these labels will need to be added with individual kustomize patches similar to https://github.com/knative/eventing-sources/blob/6fde0bf31f27dc744cd4fd59b7ff7f0c47b62f4c/config/default/containersource_conditions_patch.yaml. There's probably a way to apply one patch to all CRDs, but I'm not sure how to do it.

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: containersources.sources.eventing.knative.dev
  labels:
    eventing.knative.dev/source: "true"
```

vaikas-google:
Thanks! I'll take a look at how these are created!

PR #185, opened 2019-01-18
Label source CRDs eventing.knative.dev/source=true

### LINKED ISSUE #176: make it easier to see installed Source CRs in a cluster.

grantr:
Supersedes #183.
Fixes #176.

@vaikas-google I think you said you were having trouble with #183 so I thought I'd give it a try; hope you don't mind :)

Sources are using controller-tools to generate their CRD yamls via annotations, but controller-tools doesn't support labeling CRDs yet. Use kustomize patches to add labels instead. There's probably a prettier way to do this but this was the quickest.

## Proposed Changes

  * Add label eventing.knative.dev/source: "true"

**Release Note**

```release-note
Each source CRD is now labeled with:
    eventing.knative.dev/source: "true"
which means you can find list of available sources with:
 kubectl get crds -l "eventing.knative.dev/source=true"
```

### DIFF
--- a/config/default/kuberneteseventsource_label_patch.yaml
+++ b/config/default/kuberneteseventsource_label_patch.yaml
@@ -1,0 +1,6 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: kuberneteseventsources.sources.eventing.knative.dev
+  labels:
+    eventing.knative.dev/source: "true"
--- a/config/awssqs/kustomization.yaml
+++ b/config/awssqs/kustomization.yaml
@@ -16,4 +16,5 @@ resources:
 
 patches:
   - conditions_patch.yaml
+  - label_patch.yaml
   - env_patch.yaml
--- a/config/awssqs/label_patch.yaml
+++ b/config/awssqs/label_patch.yaml
@@ -1,0 +1,6 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: awssqssources.sources.eventing.knative.dev
+  labels:
+    eventing.knative.dev/source: "true"
--- a/config/default-awssqs.yaml
+++ b/config/default-awssqs.yaml
@@ -11,6 +11,7 @@ metadata:
   creationTimestamp: null
   labels:
     controller-tools.k8s.io: "1.0"
+    eventing.knative.dev/source: "true"
   name: awssqssources.sources.eventing.knative.dev
 spec:
   group: sources.eventing.knative.dev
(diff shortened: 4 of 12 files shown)


grantr:
I'm not sure why the build fails verifying manifests - it works for me with the latest kustomize (1.0.11).

vaikas-google:
@grantr Thanks. Yeah, as you can see in #183 I added something that looks pretty much identical to your contrib gcp pubsub and it was failing and it wasn't clear to me why, so went' to do something else :) Guess, there might be something else wrong besides just me :)

evankanderson:
Might we have picked up a new version of `kustomize`?

grantr:
We did, but it's an unreleased version: https://github.com/kubernetes-sigs/kustomize/pull/618.

The test-infra library has a feature that can get go tools and run them without being vendored, but it always uses master. I've got a branch in progress that vendors kustomize so we can constrain it to a specific release both locally and in CI. We already do this for controller-tools at https://github.com/knative/eventing-sources/blob/518626c9b16f6ad45aef00577fbcf3e249e1a584/hack/update-manifests.sh#L31.

config/default/kuberneteseventsource_label_patch.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: kuberneteseventsources.sources.eventing.knative.dev
+  labels:
+    eventing.knative.dev/source: "true"
```
matzew: Nice seeing this generic label!

grantr:
Superseded by #192

PR #186, opened 2019-01-18
Record more detail about ContainerSource reconcile

grantr:
Adds some new events and more details in condition messages to make the ContainerSource controller's actions easier to inspect and debug. The condition reason `Deploying` was changed to `DeploymentCreated` to better describe the state of the child deployment object.

Here's the list of events recorded for a ContainerSource:

Name|Type|Description
|---|---|---|
SetSinkURIFailed | Warning | Couldn't retrieve the sink URI. This should catch most of the misconfiguration errors. Includes the go error string.|
DeploymentGetFailed | Warning | Couldn't retrieve the Deployment. Includes the go error string.
DeploymentCreateFailed | Warning | Error creating the Deployment. Includes the go error string.
DeploymentCreated | Normal | A deployment was created successfully.
DeploymentUpdated | Normal | The deployment's spec was updated to the expected value.
DeploymentUpdateFailed | Warning | Error updating the Deployment. Includes the go error string.
DeploymentReady|  Normal | The deployment has at least 1 ready replica. This is the "Success" event.

When an event is recorded, the relevant condition is also updated with the same message.

See https://github.com/knative/eventing/issues/724.

## Proposed Changes

  * Record 3 new events during ContainerSource reconcile.
  * Clarify the deployment created condition reason.
  * Add go error strings to several condition and event messages.

**Release Note**

```release-note
ContainerSource condition and event messages are now more detailed when errors occur.
```


pkg/controller/containersource/reconcile.go
```diff
@@ -91,16 +92,20 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 		if errors.IsNotFound(err) {
 			deploy, err = r.createDeployment(ctx, source, nil, args)
 			if err != nil {
-				r.recorder.Eventf(source, corev1.EventTypeNormal, "DeploymentBlocked", "waiting for %v", err)
+				r.recorder.Eventf(source, corev1.EventTypeWarning, "DeploymentCreateFailed", "Could not create deployment: %v", err)
```
Harwayne: Is it worth creating a method that both emits an event and updates the status?

```golang
func (r *reconciler) markStatusDeployingAndEmitEvent(source, eventType, key, message, args) {}

...

markStatusDeployingAndEmitEvent(source, corev1.EventTypeWarning, "DeploymentCreateFailed", "Could not create deployment: %v", err)
```

pkg/controller/containersource/reconcile.go
```diff
@@ -83,6 +83,7 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) (runt
 
 	err = r.setSinkURIArg(ctx, source, args)
 	if err != nil {
+		r.recorder.Eventf(source, corev1.EventTypeWarning, "SetSinkURIFailed", "Failed to set Sink URI: %v", err)
```
Harwayne: While this works, everywhere else the status and event are set/emitted one right after the other. Why is this one separated out?

pkg/apis/sources/v1alpha1/containersource_types.go
```diff
@@ -117,6 +117,16 @@ func (s *ContainerSourceStatus) MarkNoSink(reason, messageFormat string, message
 	containerCondSet.Manage(s).MarkFalse(ContainerConditionSinkProvided, reason, messageFormat, messageA...)
 }
 
+// IsDeployed returns true if the Deployed condition has status true, otherwise
+// false.
+func (s *ContainerSourceStatus) IsDeployed() bool {
+	c := containerCondSet.Manage(s).GetCondition(ContainerConditionDeployed)
+	if c == nil || !c.IsTrue() {
```
Harwayne: I think the negation of `IsTrue()` rubs me the wrong way.

How about? 

```golang
if c != nil {
  return c.IsTrue();
}
return false
```

pkg/apis/sources/v1alpha1/containersource_types.go
```diff
@@ -117,6 +117,16 @@ func (s *ContainerSourceStatus) MarkNoSink(reason, messageFormat string, message
 	containerCondSet.Manage(s).MarkFalse(ContainerConditionSinkProvided, reason, messageFormat, messageA...)
 }
 
+// IsDeployed returns true if the Deployed condition has status true, otherwise
+// false.
+func (s *ContainerSourceStatus) IsDeployed() bool {
```
Harwayne: Unit test. :stuck_out_tongue:

pkg/apis/sources/v1alpha1/containersource_types.go
grantr: I copied that from [the `IsHappy` method from knative/pkg](https://github.com/grantr/eventing-sources/blob/04b2f641a1e255b736cd4abae00553c12ac3b49f/vendor/github.com/knative/pkg/apis/duck/v1alpha1/condition_set.go#L164-L169). Not sure why it's written that way there, but I agree yours is better.

pkg/apis/sources/v1alpha1/containersource_types.go
grantr: Done.

pkg/controller/containersource/reconcile.go
Harwayne: It's fine as-is. Maybe add a comment in `setSinkURIArg` saying that an event will be recorded outside if an error is returned.

pkg/controller/containersource/reconcile.go
Harwayne: LGTM

Harwayne:
/lgtm
/hold
Hold so you have a chance to do the optional comments.

pkg/controller/containersource/reconcile.go
grantr: Done.

grantr:
RFAL
/hold cancel

n3wscott:
I want this! gimme! :D

PR #196, opened 2019-01-31
Update test-infra

adrcunha:
Major change: use standard presubmit test runners.

Last PR that fixes knative/test-infra#170


test/e2e-tests.sh
```diff
@@ -29,8 +29,7 @@ source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/e2e-tests.s
 
 # Script entry point.
 
-initialize $@
-
-# TODO(#8): Write E2E tests.
+echo "TODO(#8): Write E2E tests."
+# initialize $@
```
evankanderson: Is this `initialize` still valid once a test is present? Is it harmful if we call `initialize $@` with no e2e tests at the moment?

adrcunha: It's not harmful, but you're wasting time and resources creating a cluster that's not used. I'll add a comment clarifying.

adrcunha: Actually I just removed the commented code.

adrcunha:
> I'm presuming this will still create the releases from the files in `contrib/`?

The behavior of `release.sh` didn't change, just its format.

PR #197, opened 2019-01-31
Change Signature of KnativeReconciler.Reconcile

n3wscott:
Fixing a minor bug with the reconciler editing the cache and changing the signature of Reconcile to make it more clear this was a bug.

## Proposed Changes

  * Change Signature of KnativeReconciler.Reconcile to `Reconcile(ctx context.Context, object runtime.Object) error`

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controller/sdk/reconciler.go
+++ b/pkg/controller/sdk/reconciler.go
@@ -65,6 +65,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, err
 	}
 
+	// Don't modify the cache's copy
 	obj := original.DeepCopyObject()
 
 	// Reconcile this copy of the Source and then write back any status
--- a/pkg/controller/sdk/reconciler.go
+++ b/pkg/controller/sdk/reconciler.go
@@ -51,9 +51,9 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	logger.Infof("Reconciling %s %v", r.provider.Parent.GetObjectKind(), request)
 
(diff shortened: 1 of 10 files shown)


grantr:
@n3wscott can you describe what the bug was? I couldn't discern it by looking at the changes.

pkg/controller/sdk/reconciler.go
```diff
@@ -51,9 +51,9 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 
 	logger.Infof("Reconciling %s %v", r.provider.Parent.GetObjectKind(), request)
 
-	obj := r.provider.Parent.DeepCopyObject()
+	original := r.provider.Parent.DeepCopyObject()
```
n3wscott: This was the bug @grantr 

I was passing in the cache's version of the object by mistake. And then assigning the returned value back to the pointer but those point to the caches version all the time....

pkg/controller/sdk/reconciler.go
```diff
@@ -65,11 +65,12 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, err
 	}
 
-	original := obj.DeepCopyObject()
+	// Don't modify the cache's copy
+	obj := original.DeepCopyObject()
```
grantr: Optional: Can we call this something like `reconciledObj` or `updated` to make clear that this holds the updated state?

grantr:
Oops, my hold didn't work. Well if you feel like taking my optional name change suggestion @n3wscott, you can create a new PR :)

PR #211, opened 2019-02-19
Add event source CloudEventIngress

### LINKED ISSUE #151: Create a CloudEvents Ingress source
A useful generic Source would be a Source which sets up routing from the external world to an `Addressable` (which may have an address which only exists within the cluster). This would assume that the incoming data was in CloudEvents format, and might be able to perform authentication and credential exchange within the cluster.

Implementation-wise, this could be another event source which using Knative Serving, or could be direct configuration of Istio/whatnot. (The advantage of using Serving is that Serving already needs to solve the ingress/TLS/multi-routing problems.)

Reasons to make a generic Source rather than simply use Knative Serving/whatnot directly:

1. We can use this to start applying policy (e.g. Auth)
3. Allows you to deliver external CloudEvents to a Channel.
2. Better capture the flow of events.

daisy-ycguo:
Fixes #151

## Proposed Changes

  * Add CloudEventIngress CRD for managing this event source
  * A Knative Service is created for every CloudEventIngress, waiting for CloudEvent message and forwarding to a sink URI.

**Release Note**

```release-note
```


cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,67 @@
+/*
+Copyright 2018 The Knative Authors
```
vaikas-google: could you change this to 2019 please :)

pkg/adapter/cloudeventingress/adapter.go
```diff
@@ -0,0 +1,99 @@
...
+	"net/http"
+	"time"
+
+	"github.com/knative/pkg/cloudevents"
+	"github.com/knative/pkg/logging"
+	"go.uber.org/zap"
+)
+
+const (
+	eventType = "dev.knative.cronjob.event"
```
vaikas-google: is this right?

pkg/adapter/cloudeventingress/adapter.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"net/http"
+	"time"
+
+	"github.com/knative/pkg/cloudevents"
```
vaikas-google: With the plan to move to using cloud-events sdk, perhaps consider using that from the get go.
https://github.com/knative/eventing-sources/pull/214

Another thought is that with the proposed Broker/Trigger model, this ingress might actually make more sense to live in the Eventing repo? But haven't reviewed all, so perhaps I'm wrong.

pkg/adapter/cloudeventingress/adapter.go
```diff
@@ -0,0 +1,99 @@
...
+// Adapter implements the CloudEventsIngress to trigger a Sink.
+type Adapter struct {
+	// Messasge is the data to be posted to the target.
+	Message string
+
+	// SinkURI is the URI messages will be forwarded on to.
+	SinkURI string
+}
+
+func (a *Adapter) Start(ctx context.Context) error {
```
Harwayne: This looks like it sends a single event once, is that what it is supposed to be doing?

pkg/reconciler/cloudeventingress/resources/cloudeventingress.go
```diff
@@ -0,0 +1,81 @@
...
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CloudEventIngressArgs are the arguments needed to create a CloudEvent Ingress. Every
+// field is required.
+type CloudEventIngressArgs struct {
+	Image   string
+	Source  *v1alpha1.CronJobSource
```
dubee: Is this `Source` correct?

dubee: Should be `CloudEventIngress`, I believe.

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,74 @@
...
+	flag.StringVar(&sink, "sink", "", "uri to send events to")
+}
+
+func receive(event cloudevents.Event) {
+	ec := event.Context.AsV02()
+	log.Printf("CloudEvent:\n%s", event)
+	log.Printf("[%s] %s %s: ", ec.Time, event.DataContentType(), ec.Source.String())
+
+	//send out
+	if _, err := ceClient.Send(context.TODO(), event); err != nil {
```
n3wscott: if you change your receive method to be: `func receive(ctx context.Context, event cloudevents.Event)` you can forward on the context to `Send`.

pkg/apis/sources/v1alpha1/containersource_types.go
```diff
@@ -24,6 +24,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// Important: Run "make" to regenerate code after modifying this file
```
n3wscott: this comment is no longer valid, we have no make file.

pkg/reconciler/cloudeventingress/cloudeventingress.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		logger.Error("Unable to create the receive adapter", zap.Error(err))
+		return err
+	}
+	src.Status.MarkDeployed()
+	return nil
+}
+
+func (r *reconciler) createKnativeService(ctx context.Context, src *v1alpha1.CloudEventIngress, sinkURI string) error {
+	// ra, err := r.getReceiveAdapter(ctx, src)
```
n3wscott: need to clean up the commented out code.

pkg/reconciler/cloudeventingress/cloudeventingress.go
```diff
@@ -0,0 +1,148 @@
...
+	// 	logging.FromContext(ctx).Error("Unable to get an existing receive adapter", zap.Error(err))
+	// 	return nil, err
+	// }
+	args := resources.CloudEventIngressArgs{
+		Image:   r.receiveAdapterImage,
+		Source:  src,
+		SinkURI: sinkURI,
+	}
+	expected := resources.MakeCloudEventIngress(&args)
+	// if ra != nil {
```
n3wscott: and here

pkg/reconciler/cloudeventingress/doc.go
```diff
@@ -0,0 +1,19 @@
+/*
+Copyright 2018 The Knative Authors
```
n3wscott: 2019

pkg/apis/sources/v1alpha1/cloudeventingress_types.go
```diff
@@ -0,0 +1,137 @@
...
+
+package v1alpha1
+
+import (
+	duckv1alpha1 "github.com/knative/pkg/apis/duck/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
n3wscott: delete kubebuilder comments please

pkg/apis/sources/v1alpha1/cloudeventingress_types.go
```diff
@@ -0,0 +1,137 @@
...
+	// Sink is the URI
+	Sink *corev1.ObjectReference `json:"sink,omitempty"`
+	// ServiceAccountName is the name of the ServiceAccount to use to run this
+	// source.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+}
+
+// CloudEventIngressStatus defines the observed state of CloudEventIngress
+type CloudEventIngressStatus struct {
```
n3wscott: Please see https://github.com/knative/eventing-sources/pull/258 for migrating to the ducktype for embedded status.

pkg/reconciler/cloudeventingress/cloudeventingress.go
```diff
@@ -0,0 +1,148 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+const (
+	// controllerAgentName is the string used by this controller to identify
+	// itself when creating events.
+	controllerAgentName = "cloudeventingress-controller"
```
n3wscott: I think it should be `cloudeventsingress-controller`

This comment applies to anywhere `cloudevent` is used. I think the only valid term is `cloudevents`.

pkg/reconciler/cloudeventingress/resources/service.go
```diff
@@ -0,0 +1,68 @@
+/*
+Copyright 2018 The Knative Authors
```
n3wscott: 2019

pkg/reconciler/cloudeventingress/resources/service.go
```diff
@@ -0,0 +1,68 @@
...
+// MakeCloudEventIngress generates (but does not insert into K8s) the CloudEventIngress Deployment
+// and Service for CloudEvent Sources.
+func MakeCloudEventIngress(args *CloudEventIngressArgs) *sv1alpha1.Service {
+	labels := map[string]string{
+		"knative-eventing-source": "cloudeventingress",
+	}
+	return &sv1alpha1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:    args.Source.Namespace,
+			GenerateName: fmt.Sprintf("cloudeventingress-%s-", args.Source.Name),
```
n3wscott: cloudevents

pkg/reconciler/cloudeventingress/resources/service.go
```diff
@@ -0,0 +1,68 @@
...
+// field is required.
+type CloudEventIngressArgs struct {
+	Image   string
+	Source  *v1alpha1.CloudEventIngress
+	SinkURI string
+}
+
+// MakeCloudEventIngress generates (but does not insert into K8s) the CloudEventIngress Deployment
+// and Service for CloudEvent Sources.
+func MakeCloudEventIngress(args *CloudEventIngressArgs) *sv1alpha1.Service {
```
n3wscott: CloudEvents

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,74 @@
...
+func main() {
+	ctx := context.TODO()
+
+	flag.Parse()
+
+	if sink == "" {
+		log.Fatalf("no sink provided")
+	}
+
+	ceClient, err := kncloudevents.NewDefaultClient()
```
n3wscott: The client has no sink.

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,74 @@
...
+	"log"
+
+	"github.com/cloudevents/sdk-go/pkg/cloudevents"
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/client"
+	"github.com/knative/eventing-sources/pkg/kncloudevents"
+)
+
+const (
+	// Sink to transform CloudEvent data.
+	envSinkURI = "SINK_URI"
```
n3wscott: This is never used.

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,74 @@
...
+	if sink == "" {
+		log.Fatalf("no sink provided")
+	}
+
+	ceClient, err := kncloudevents.NewDefaultClient()
+	if err != nil {
+		log.Fatalf("failed to create client:%s", err)
+	}
+
+	if err := ceClient.StartReceiver(ctx, receive); err != nil {
```
n3wscott: StartReceiver is now blocking in CloudEvents SDK >=0.4.0 so you can just do this now:

```
log.Fatalf(ceClient.StartReceiver(ctx, receive))
```

pkg/reconciler/cloudeventingress/resources/service.go
```diff
@@ -0,0 +1,68 @@
...
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"fmt"
+
+	"github.com/knative/eventing-sources/pkg/apis/sources/v1alpha1"
+	sv1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
```
n3wscott: the standard way to import this would be `servingv1alpha1`, more wordy but very clear what it brings.

pkg/reconciler/cloudeventingress/resources/service.go
```diff
@@ -0,0 +1,68 @@
...
+			RunLatest: &sv1alpha1.RunLatestType{
+				Configuration: sv1alpha1.ConfigurationSpec{
+					RevisionTemplate: sv1alpha1.RevisionTemplateSpec{
+						Spec: sv1alpha1.RevisionSpec{
+							Container: corev1.Container{
+								Image: args.Image,
+								Env: []corev1.EnvVar{
+									{
+										Name:  "SINK_URI",
+										Value: args.SinkURI,
```
n3wscott: `SINK_URI` is never used.

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,74 @@
...
+	envSinkURI = "SINK_URI"
+)
+
+var (
+	sink     string
+	ceClient client.Client
+)
+
+func init() {
+	flag.StringVar(&sink, "sink", "", "uri to send events to")
```
n3wscott: after doing these types of things a couple times, my advice is only allow one way to configure the source, just envargs is easiest for pods. and not hard locally:

```
SINK_URI=http://somehost.com/woot go run cmd/my/tool/main.go
```

cmd/cloudeventingress/main.go
daisy-ycguo: it used in the latest code base: sink := os.Getenv(envSinkURI)

pkg/reconciler/cloudeventingress/cloudeventingress.go
```diff
@@ -0,0 +1,165 @@
...
+func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) error {
+	logger := logging.FromContext(ctx)
+
+	src, ok := object.(*v1alpha1.CloudEventIngress)
+	if !ok {
+		logger.Error("could not find CloudEventIngress source", zap.Any("object", object))
+		return nil
+	}
+
+	// This Source attempts to reconcile three things.
```
duglin: 3 or 2? :-)

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,67 @@
...
+	ceClient client.Client
+)
+
+func receive(ctx context.Context, event cloudevents.Event) {
+	ec := event.Context.AsV02()
+	log.Printf("CloudEvent:\n%s", event)
+	log.Printf("[%s] %s %s: ", ec.Time, event.DataContentType(), ec.Source.String())
+
+	//send out
+	if _, err := ceClient.Send(context.TODO(), event); err != nil {
```
n3wscott: you should send the context that was provided in receive.

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,67 @@
...
+		log.Fatalf("failed to send cloudevent\n%s", err)
+	}
+}
+
+func main() {
+	ctx := context.TODO()
```
n3wscott: use `context.Background()`, not `TODO()`

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,67 @@
...
+
+	sink := os.Getenv(envSinkURI)
```
n3wscott: sink is still not used from what I can see.

you can do this:

```
ctx = cecontext.WithTarget(ctx, sink)
```

daisy-ycguo:
I think we could discuss whether to use a ContainerService or a Knative Service. I prefer a Knative Service because it can leverage the existing mechanism, e.g. the scalability, the security, the ingress, and etc.

@Harwayne  @vaikas-google here need your inputs too. Thanks.

cmd/cloudeventingress/main.go
daisy-ycguo: Done

pkg/adapter/cloudeventingress/adapter.go
daisy-ycguo: Changed. Thanks.

pkg/adapter/cloudeventingress/adapter.go
daisy-ycguo: cloud-events sdk is used.

pkg/adapter/cloudeventingress/adapter.go
daisy-ycguo: Changed. Please review the new version.

pkg/reconciler/cloudeventingress/resources/cloudeventingress.go
daisy-ycguo: Changed. Please review the new version.

daisy-ycguo: Changed. Please review the new version.

cmd/cloudeventingress/main.go
daisy-ycguo: Changed.

pkg/apis/sources/v1alpha1/containersource_types.go
daisy-ycguo: Comment removed

pkg/reconciler/cloudeventingress/cloudeventingress.go
daisy-ycguo: Comment removed.

pkg/reconciler/cloudeventingress/cloudeventingress.go
daisy-ycguo: Comment removed.

pkg/reconciler/cloudeventingress/doc.go
daisy-ycguo: Done.

pkg/apis/sources/v1alpha1/cloudeventingress_types.go
daisy-ycguo: Comment removed.

pkg/apis/sources/v1alpha1/cloudeventingress_types.go
daisy-ycguo: Updated. Please review the new version.

pkg/reconciler/cloudeventingress/resources/service.go
daisy-ycguo: Updated.

pkg/reconciler/cloudeventingress/resources/service.go
daisy-ycguo: Updated. Please review the new version.

cmd/cloudeventingress/main.go
n3wscott: You need to also use the context object. 

```
if _, err := ceClient.Send(context.TODO(), event); err != nil {
```

becomes 

```
if _, err := ceClient.Send(ctx, event); err != nil {
```

cmd/cloudeventingress/main.go
n3wscott: ping

pkg/reconciler/cloudeventingress/resources/service.go
n3wscott: now you have `sink := os.Getenv(envSinkURI)` but `sink` is never used.

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func receive(ctx context.Context, event cloudevents.Event) {
+	ec := event.Context.AsV02()
+	log.Printf("CloudEvent:\n%s", event)
+	log.Printf("[%s] %s %s: ", ec.Time, event.DataContentType(), ec.Source.String())
+
+	//send out
+	if _, err := ceClient.Send(context.TODO(), event); err != nil {
+		log.Fatalf("failed to send cloudevent\n%s", err)
```
Harwayne: This should probably be `log.Errorf` instead, so that the entire process doesn't go down if downstream has a hiccup and rejects a request.

pkg/apis/sources/v1alpha1/cloudeventingress_types.go
```diff
@@ -0,0 +1,140 @@
...
+// Check that KubernetesEventSource can be validated and can be defaulted.
+var _ runtime.Object = (*CloudEventIngress)(nil)
+
+// Check that KubernetesEventSource implements the Conditions duck type.
+var _ = duck.VerifyType(&CloudEventIngress{}, &duckv1alpha1.Conditions{})
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// CloudEventIngress is the Schema for the cloudeventingresses API
```
Harwayne: Add a period at the end to make this a complete sentence.

pkg/apis/sources/v1alpha1/cloudeventingress_types.go
```diff
@@ -0,0 +1,140 @@
...
+	cloudEventIngressCondSet.Manage(s).InitializeConditions()
+}
+
+// MarkSink sets the condition that the source has a sink configured.
+func (s *CloudEventIngressStatus) MarkSink(uri string) {
+	s.SinkURI = uri
+	if len(uri) > 0 {
+		cloudEventIngressCondSet.Manage(s).MarkTrue(CloudEventIngressConditionSinkProvided)
+	} else {
+		cloudEventIngressCondSet.Manage(s).MarkUnknown(CloudEventIngressConditionSinkProvided, "SinkEmpty", "Sink has resolved to empty.%s", "")
```
Harwayne: Is the empty string needed? Would the following work?

```golang
cloudEventIngressCondSet.Manage(s).MarkUnknown(CloudEventIngressConditionSinkProvided, "SinkEmpty", "Sink has resolved to empty.")
```

pkg/reconciler/cloudeventingress/cloudeventingress.go
Harwayne: This is fairly pervasive. Does this also affect the name of packages and the CRD?

pkg/reconciler/cloudeventingress/cloudeventingress.go
```diff
@@ -0,0 +1,165 @@
...
+			if err != nil {
+				logger.Error("Unable to create the receive adapter", zap.Error(err))
+				return err
+			}
+			src.Status.MarkDeployed()
+			return nil
+		}
+		return err
+	}
+
```
Harwayne: We should probably add `src.Status.MarkDeployed()` here too. Because if we don't, then it is possible for us to not ever record that status change. e.g. something else modified the object concurrently, so the Status wasn't written, on the next reconcile loop, `r.getOwnedService` would return it, so we wouldn't hit the create path again.

Harwayne: We need to record the address of the created Knative Service in the CloudEventsIngress' status. As it is, the Ingress is created, but we don't know what its name is (and therefore where to send events) without inspecting the Knative Service itself.

pkg/reconciler/cloudeventingress/resources/service.go
```diff
@@ -0,0 +1,69 @@
...
+// MakeCloudEventIngress generates (but does not insert into K8s) the CloudEventIngress Deployment
+// and Service for CloudEvent Sources.
+func MakeCloudEventIngress(args *CloudEventIngressArgs) *servingv1alpha1.Service {
+	labels := map[string]string{
+		"knative-eventing-source": "cloudeventingress",
+	}
+
+	return &servingv1alpha1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("%s-", args.Source.Name),
```
Harwayne: We probably shouldn't use GenerateName. I think that people will want to have exact control over the name of the ingress' URL. Instead we should use something like `Name: fmt.Sprintf("%s-ce-ingress", args.Source.Name)`

config/300-cloudeventingress.yaml
```diff
@@ -0,0 +1,77 @@
...
+  name: cloudeventingresses.sources.eventing.knative.dev
+spec:
+  group: sources.eventing.knative.dev
+  names:
+    categories:
+    - all
+    - knative
+    - eventing
+    - sources
+    kind: CloudEventIngress
```
Harwayne: Every other existing source ends in 'source'. Should this one to? @vaikas-google E.g. `CloudEventsIngressSource` (@n3wscott already mentioned that CloudEvent -> CloudEvents).

pkg/reconciler/cloudeventingress/cloudeventingress.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	logging.FromContext(ctx).Info("CloudEventIngress created. %s", expected)
+	return nil
+}
+
+func (r *reconciler) getOwnedService(ctx context.Context, source *v1alpha1.CloudEventIngress) (*servingv1alpha1.Service, error) {
+	list := &servingv1alpha1.ServiceList{}
+	err := r.client.List(ctx, &client.ListOptions{
+		Namespace:     source.Namespace,
+		LabelSelector: labels.Everything(),
```
Harwayne: Make a label selector with the labels you use when creating the Knative Service.

pkg/reconciler/cloudeventingress/resources/service.go
```diff
@@ -0,0 +1,69 @@
...
+	Image   string
+	Source  *v1alpha1.CloudEventIngress
+	SinkURI string
+}
+
+// MakeCloudEventIngress generates (but does not insert into K8s) the CloudEventIngress Deployment
+// and Service for CloudEvent Sources.
+func MakeCloudEventIngress(args *CloudEventIngressArgs) *servingv1alpha1.Service {
+	labels := map[string]string{
+		"knative-eventing-source": "cloudeventingress",
```
Harwayne: Add a second label with the name of the CloudEventsIngressSource, something like:

```golang
"sources.eventing.knative.dev/cloudEventsIngressSource": args.Source.Name,
```

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,67 @@
...
+var (
+	sink     string
+	ceClient client.Client
+)
+
+func receive(ctx context.Context, event cloudevents.Event) {
+	ec := event.Context.AsV02()
+	log.Printf("CloudEvent:\n%s", event)
+	log.Printf("[%s] %s %s: ", ec.Time, event.DataContentType(), ec.Source.String())
+
```
Harwayne: As mentioned by @n3wscott below, you need to provide the ceClient the sink's address.

```golang
ctx = cecontext.WithTarget(ctx, sink)
```

cmd/cloudeventingress/main.go
```diff
@@ -0,0 +1,67 @@
...
+	ceClient, err := kncloudevents.NewDefaultClient()
+	if err != nil {
+		log.Fatalf("failed to create client:%s", err)
+	}
+
+	log.Fatalf("failed to start receiver:%s", ceClient.StartReceiver(ctx, receive))
+
+	log.Printf("listening on port %d\n", 8080)
+	<-ctx.Done()
```
Harwayne: I don't think this will ever return, is that the intention?

pkg/reconciler/cloudeventingress/cloudeventingress.go
daisy-ycguo: Fixed. Thanks.

cmd/cloudeventingress/main.go
daisy-ycguo: Fixed. Thanks.

cmd/cloudeventingress/main.go
daisy-ycguo: fixed. Thanks.

cmd/cloudeventingress/main.go
daisy-ycguo: fixed. Thanks.

pkg/apis/sources/v1alpha1/cloudeventingress_types.go
daisy-ycguo: Fixed. Thanks.

cmd/cloudeventingress/main.go
daisy-ycguo: Change to `Printf` cause no `Errorf` defined.

cmd/cloudeventingress/main.go
daisy-ycguo: Fixed. Thanks.

config/300-cloudeventingress.yaml
daisy-ycguo: Agree to add a `Source` ending.

daisy-ycguo:
Close this pull request because a lot of changes have happened. A new PR will be submitted later.

PR #214, opened 2019-02-21
updating heartbeat sample to show the new cloudevents sdk.

n3wscott:
I have been tackling the effort of converging on a single cloudevents library. The result of that effort is an [alpha sdk](https://github.com/cloudevents/sdk-go/). The plan is to remove knative/pkg/cloudevents as I determine the new sdk fulfills our needs. 

## Proposed Changes

  * Show an example of new cloudevents sdk integration.
  * Vendor in the cloudevents lib.

**Release Note**

```release-note
NONE.
```


samples/heartbeats/README.md
```diff
@@ -0,0 +1,19 @@
+# Heartbeats Sample
+
+This is a sample showing ContainerSource
+
+## Deployment Steps
+
+### Prerequisites
+
+- [ ] Install Knative Core
```
vaikas-google: maybe a link?

samples/heartbeats/README.md
```diff
@@ -0,0 +1,19 @@
...
+- [ ] Knative Eventing Sources (this repo)
+
+### Deploy Heartbeats Sender and Receiver
+
+```bash
+ko apply -f ./samples/heartbeats/
```
vaikas-google: Could we publish an image instead of relying on ko? or, update the Prerequisites.

samples/heartbeats/README.md
n3wscott: Added a link.

samples/heartbeats/README.md
n3wscott: I was intending this for a developer thinking of implementing a source vs someone kicking the knative tires. I added a little to the top to make this more clear.

cmd/heartbeats/main.go
```diff
@@ -43,13 +50,37 @@ func init() {
...
-		EventType: "dev.knative.source.heartbeats",
-		Source:    "heartbeats-demo",
-	})
+	var env envConfig
+	if err := envconfig.Process("", &env); err != nil {
+		log.Printf("[ERROR] Failed to process env var: %s", err)
+		os.Exit(1)
+	}
+
+	if env.Sink != "" {
```
vaikas-google: Just a question if the Process above will actually fail if POD_NAME, POD_NAMESPACE are not specified? Since Sink can be empty, can Name / Namespace be too? it seems like they should fail if not specified?  Should you tag them with required?

cmd/heartbeats/main.go
```diff
@@ -43,13 +50,37 @@ func init() {
...
+	// Sink URL where to send heartbeat cloudevents
+	Sink string `envconfig:"SINK"`
+
+	// Name of this pod.
+	Name string `envconfig:"POD_NAME"`
+
+	// Namespace this pod exists in.
+	Namespace string `envconfig:"POD_NAMESPACE"`
+
+	// The ip address of this pod.
```
vaikas-google: is this used anywhere?

n3wscott: It is not, but I was so excited about finding that k8s could do this I put it in here so it can be known... I can take it out :(

cmd/heartbeats/main.go
n3wscott: it will not, they will just default to `""` but that will not be possable using the sample yaml.

I am trying to show that the source url changes per instance of heartbeat.

n3wscott: marked required.

vaikas-google: Hard to understand from the documents if that's the way you specify two things. I trust you it works :)

PR #215, opened 2019-02-21
Update k8s source to new version of cloudevents sdk

n3wscott:
A followup to https://github.com/knative/eventing-sources/pull/214

## Proposed Changes

  * Migrate k8s source to use cloudevents/sdk-go

**Release Note**

```release-note
K8s source now producing version 0.2 CloudEvents.
```


pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -101,11 +99,16 @@ func (a *Adapter) addEvent(new interface{}) {
 	logger := logging.FromContext(context.TODO()).With(zap.Any("eventID", event.ObjectMeta.UID), zap.Any("sink", a.SinkURI))
 	logger.Debug("GOT EVENT", event)
 
-	if a.client == nil {
-		a.client = cloudevents.NewClient(a.SinkURI, cloudevents.Builder{EventType: eventType})
+	if a.ceClient == nil {
+		var err error
+		a.ceClient, err = client.NewHttpClient(context.TODO(), a.SinkURI, http.BinaryV02)
```
evankanderson: Can any of the arguments default?

context seems like it should be `context.Background()`, since you're not forwarding the context from elsewhere, you're acting on your own authority.

pkg/adapter/kubernetesevents/cloudevent.go
```diff
@@ -75,10 +81,14 @@ func createSelfLink(o corev1.ObjectReference) string {
...
 //		EventTime:0001-01-01 00:00:00 +0000 UTC,
 //	}
-func cloudEventOverrides(m *corev1.Event) cloudevents.V01EventContext {
-	return cloudevents.V01EventContext{
-		EventID:   string(m.ObjectMeta.UID),
-		Source:    createSelfLink(m.InvolvedObject),
-		EventTime: m.ObjectMeta.CreationTimestamp.Time,
+func cloudEventFrom(m *corev1.Event) cloudevents.Event {
+	return cloudevents.Event{
+		Context: cloudevents.EventContextV01{
```
evankanderson: It would be nice to have a factory here which could default the EventID (and possibly time, though you'd want to override that here).

Do you want this to be a CE V0.1 context, or use the V0.2 names here?

n3wscott: The new lib does not care what you give it, only what the transport is set to. So this results in the same thing. Maybe magic, maybe neat?

evankanderson: I know it results in the same thing. 😉

 I'm wondering if you want to use the newer names so that newer readers aren't too confused. 😁

n3wscott: got it, yes! Done.

pkg/adapter/kubernetesevents/adapter.go
n3wscott: HECK YEAH NOW THEY CAN!!!

n3wscott:
waiting to rebase after https://github.com/knative/eventing-sources/pull/222

n3wscott:
/hold cancel

Updated. PTAL. Thanks!!

pkg/adapter/kubernetesevents/cloudevent_test.go
```diff
@@ -55,10 +55,11 @@ func TestCloudEventFrom(t *testing.T) {
 
 	want := cloudevents.Event{
 		Context: cloudevents.EventContextV02{
-			Type:   eventType,
-			ID:     string(uid),
-			Source: *cetypes.ParseURLRef(refLink),
-			Time:   &cetypes.Timestamp{Time: now},
+			SpecVersion: "0.2",
```
vaikas-google: No need for action, just a comment that I find it counterintuitive that by using the EventContextV02 you still have to specify SpecVersion as 0.2.

vaikas-google:
/approve
leaving lgtm for @evankanderson

pkg/adapter/kubernetesevents/cloudevent_test.go
n3wscott: This is on the Want side. you don't if you ask for it as for the struct back `.AsV02()` it fill out that field for you.

vaikas-google: duh, thanks :)

Harwayne:
How will this interact with all of the existing knative/eventing pieces? I think they all currently assume Cloud Events v0.1. Do we need to update those pieces first?

evankanderson:
The current knative/eventing pieces don't assume CloudEvents at all, but do have a whitelist of fields they will forward:

https://github.com/knative/eventing/blob/master/pkg/provisioners/message_receiver.go#L164
https://github.com/knative/eventing/blob/master/pkg/provisioners/message.go#L34

n3wscott:
> How will this interact with all of the existing knative/eventing pieces? I think they all currently assume Cloud Events v0.1. Do we need to update those pieces first?

I will update everything.

> The current knative/eventing pieces don't assume CloudEvents at all, but do have a whitelist of fields they will forward:
> 
> https://github.com/knative/eventing/blob/master/pkg/provisioners/message_receiver.go#L164
> https://github.com/knative/eventing/blob/master/pkg/provisioners/message.go#L34

Hm, more stuff to update, thanks for the link!

PR #218, opened 2019-02-22
Modify github source to add ability to specify api url for GHE

dibbles:
This change enables optionally specifying a githubAPIURL on the githubsource, saving the need for people to have to create a completly new eventsource to work with github enterprise.  

The new githubAPIURL field is optional and if not specified will default to the current behaviour, namely having the github public api base URL.

## Proposed Changes

  * Introduce optional githubAPIURL field on githubsource.  Users can then use a githubsource and specify the API URL of their GitHub Enterprise server

**Release Note**
```release-note
NONE
```

Note: Doc changes ready, PR not yet open as in docs repo (please advise when reviewing if I should open a PR and comment that it cannot be merged until after this change) - this is my first contribution so not exactly clear on process with multiple repos.


dibbles:
Arrrrggghhh!!!! ..... I will get this CLA email thing correct.  Cancelling to resubmit.

PR #222, opened 2019-02-25
Updated the CloudEvent SDK, down-streaming those changes.

n3wscott:
Notable changes: HTTP casing, and new client.Option usage.

**Release Note**

```release-note
NONE
```


n3wscott:
/assign @evankanderson 

since Ville is traveling

Gopkg.toml
```diff
@@ -69,8 +69,6 @@ required = [
   name = "github.com/knative/serving"
   version = "v0.2.1"
 
-# STANZAS BELOW ARE GENERATED AND MAY BE WRITTEN - DO NOT MODIFY BELOW THIS LINE.
```
evankanderson: ???

cmd/heartbeats/main.go
```diff
@@ -110,7 +110,7 @@ func main() {
 			Data: hb,
 		}
 
-		if err := c.Send(event); err != nil {
+		if err := c.Send(context.TODO(), event); err != nil {
```
evankanderson: `context.Background()`.

Gopkg.toml
n3wscott: We do not auto-generate anymore. This line comes from kubebuilder

PR #235, opened 2019-03-07
Correct release artifacts in release.sh

### LINKED ISSUE #213: v0.4.0 gcppubsub.yaml release artifact empty
## Expected Behavior
The release artifact, `gcppubsub.yaml`, for the v0.4.0 release should contain the YAML manifest to run the GCP PubSub controller

## Actual Behavior
The release artifact contains only a comment.

grantr:
"Releases" as defined by release.sh are combinations of components. Eventing Sources releases don't have any of those.

Since components and releases share a namespace, releases cannot have
the same name as any components. This caused https://github.com/knative/eventing-sources/issues/213 as explained in https://github.com/knative/eventing-sources/issues/213#issuecomment-470654486.

Fixes #213

## Proposed Changes

  * Remove the `sources.yaml` release artifact.

**Release Note**

```release-note
The sources.yaml artifact accidentally uploaded as part of release 0.4.0 was removed.
```

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -19,17 +19,16 @@ source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/release.sh
 # Yaml files to generate, and the source config dir for them.
 declare -A COMPONENTS
 COMPONENTS=(
-  ["sources.yaml"]="config"
+  ["release.yaml"]="config"
   ["gcppubsub.yaml"]="contrib/gcppubsub/config"
   ["message-dumper.yaml"]="config/tools/message-dumper"
 )
 readonly COMPONENTS
 
+# Yaml files that are combinations of the above components. Release files must
+# not have the same name as any component files.
 declare -A RELEASES
 RELEASES=(
-  ["release.yaml"]="sources.yaml"
-  ["gcppubsub.yaml"]="gcppubsub.yaml"
-  ["message-dumper.yaml"]="message-dumper.yaml"
 )
 readonly RELEASES
 


n3wscott:
/lgtm

How will pubsub source be "released" now?

grantr:
Elements of the `COMPONENTS` map are also "released". The `RELEASES` variable is poorly named.

hack/release.sh
```diff
@@ -19,17 +19,16 @@ source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/release.sh
...
   ["gcppubsub.yaml"]="contrib/gcppubsub/config"
   ["message-dumper.yaml"]="config/tools/message-dumper"
 )
 readonly COMPONENTS
 
+# Yaml files that are combinations of the above components. Release files must
+# not have the same name as any component files.
 declare -A RELEASES
 RELEASES=(
-  ["release.yaml"]="sources.yaml"
```
adrcunha: 1. This array is empty, it can be removed.
1. Now the loop at line 44 won't be executed, but it wasn't removed from the code.

Is it intended?

adrcunha: Synced with Grant, this is intended. Dead code cleanup in https://github.com/knative/eventing-sources/pull/236

grantr: This is intended. I didn't remove the array or the loop so that this release.sh looks similar to [the eventing one](https://github.com/knative/eventing/blob/24e3c020eeb46ee18650f47ae0c297228bbfb4db/hack/release.sh).

#236 tracks this.

Issue #206, opened 2019-02-08
Data race in cronjobevents


bbrowning:
CI has caught intermittent data races in cronjobevents on both #191 and #178. I haven't done any digging into the cause yet so don't know if this is purely a race in the cronjobevents tests or a race in the non-test code as well.

Example races from a recent CI failure:
```
I0208 15:42:00.683] ==================
I0208 15:42:00.683] WARNING: DATA RACE
I0208 15:42:00.684] Read at 0x00c0000b72f0 by goroutine 55:
I0208 15:42:00.684]   github.com/knative/eventing-sources/pkg/adapter/cronjobevents.(*Adapter).cronTick()
I0208 15:42:00.685]       /go/src/github.com/knative/eventing-sources/pkg/adapter/cronjobevents/adapter.go:68 +0x97
I0208 15:42:00.685]   github.com/knative/eventing-sources/pkg/adapter/cronjobevents.(*Adapter).cronTick-fm()
I0208 15:42:00.685]       /go/src/github.com/knative/eventing-sources/pkg/adapter/cronjobevents/adapter.go:58 +0x41
I0208 15:42:00.686]   github.com/knative/eventing-sources/vendor/github.com/robfig/cron.FuncJob.Run()
I0208 15:42:00.686]       /go/src/github.com/knative/eventing-sources/vendor/github.com/robfig/cron/cron.go:92 +0x34
I0208 15:42:00.686]   github.com/knative/eventing-sources/vendor/github.com/robfig/cron.(*Cron).runWithRecovery()
I0208 15:42:00.687]       /go/src/github.com/knative/eventing-sources/vendor/github.com/robfig/cron/cron.go:165 +0x68
I0208 15:42:00.687] 
I0208 15:42:00.687] Previous write at 0x00c0000b72f0 by goroutine 22:
I0208 15:42:00.687]   github.com/knative/eventing-sources/pkg/adapter/cronjobevents.(*Adapter).cronTick()
I0208 15:42:00.688]       /go/src/github.com/knative/eventing-sources/pkg/adapter/cronjobevents/adapter.go:69 +0x309
I0208 15:42:00.688]   github.com/knative/eventing-sources/pkg/adapter/cronjobevents.TestStart_ServeHTTP.func1()
I0208 15:42:00.688]       /go/src/github.com/knative/eventing-sources/pkg/adapter/cronjobevents/adapter_test.go:72 +0x2de
I0208 15:42:00.689]   testing.tRunner()
[...]

PR #240, opened 2019-03-08
Fix race condition for creating cloudevent client

### LINKED ISSUE #227: Several sources construct their CloudEvents client without thread safety
## Expected Behavior

Sources are thread-safe.

## Actual Behavior

Several sources use variants of `if client == nil` inside the message handler function, which will race if two messages are handled concurrently.

AWSSQS source: https://github.com/knative/eventing-sources/blob/acb8625a4469cbf4969400e74ee98766d209e719/pkg/adapter/awssqs/adapter.go#L158

CronJob source: https://github.com/knative/eventing-sources/blob/acb8625a4469cbf4969400e74ee98766d209e719/pkg/adapter/cronjobevents/adapter.go#L71

GitHub source: https://github.com/knative/eventing-sources/blob/acb8625a4469cbf4969400e74ee98766d209e719/pkg/adapter/github/adapter.go#L58

## Additional Info

These client instantiations should either happen outside the message handler or they should be wrapped in a `sync.Once` (https://golang.org/pkg/sync/#Once).

/cc @n3wscott

whynowy:
Fixes #227 #206 

## Proposed Changes

  * Move client init out of PostMessage for Cronjob and AwsSqs sources
  * Use sync.Once to init client for github event source 
 

**Release Note**
No
```release-note
```

### DIFF
--- a/pkg/adapter/awssqs/adapter_test.go
+++ b/pkg/adapter/awssqs/adapter_test.go
@@ -59,8 +59,7 @@ func TestPostMessage_ServeHTTP(t *testing.T) {
 				OnFailedPollWaitSecs: 1,
 			}
 
-			err := a.initClient()
-			if err != nil {
+			if err := a.initClient(); err != nil {
 				t.Errorf("failed to create cloudevent client, %v", err)
 			}
 
@@ -74,7 +73,7 @@ func TestPostMessage_ServeHTTP(t *testing.T) {
 				Body:       &body,
 				Attributes: attrs,
 			}
-			err = a.postMessage(context.TODO(), zap.S(), m)
+			err := a.postMessage(context.TODO(), zap.S(), m)
 
 			if tc.error && err == nil {
 				t.Errorf("expected error, but got %v", err)
@@ -139,8 +138,7 @@ func TestReceiveMessage_ServeHTTP(t *testing.T) {
 				SinkURI:  sinkServer.URL,
 			}
 
-			err := a.initClient()
-			if err != nil {
+			if err := a.initClient(); err != nil {
 				t.Errorf("failed to create cloudevent client, %v", err)
 			}
 
--- a/pkg/adapter/awssqs/adapter_test.go
+++ b/pkg/adapter/awssqs/adapter_test.go
@@ -59,6 +59,11 @@ func TestPostMessage_ServeHTTP(t *testing.T) {
 				OnFailedPollWaitSecs: 1,
 			}
 
+			err := a.initClient()
+			if err != nil {
+				t.Errorf("failed to create cloudevent client, %v", err)
+			}
+
 			body := "The body"
 			messageId := "ABC01"
 			attrs := map[string]*string{
@@ -69,7 +74,7 @@ func TestPostMessage_ServeHTTP(t *testing.T) {
 				Body:       &body,
 				Attributes: attrs,
 			}
-			err := a.postMessage(context.TODO(), zap.S(), m)
+			err = a.postMessage(context.TODO(), zap.S(), m)
 
 			if tc.error && err == nil {
 				t.Errorf("expected error, but got %v", err)
@@ -132,6 +137,11 @@ func TestReceiveMessage_ServeHTTP(t *testing.T) {
 			a := &Adapter{
 				QueueURL: "https://sqs.us-east-2.amazonaws.com/123123/MyQueue",
 				SinkURI:  sinkServer.URL,
+			}
+
+			err := a.initClient()
+			if err != nil {
+				t.Errorf("failed to create cloudevent client, %v", err)
 			}
 
 			ack := new(bool)
--- a/pkg/adapter/cronjobevents/adapter.go
+++ b/pkg/adapter/cronjobevents/adapter.go
@@ -74,8 +74,7 @@ func (a *Adapter) Start(ctx context.Context, stopCh <-chan struct{}) error {
 		return err
 	}
 
-	err = a.initClient()
-	if err != nil {
+	if err = a.initClient(); err != nil {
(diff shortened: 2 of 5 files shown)


pkg/adapter/awssqs/adapter_test.go
```diff
@@ -59,6 +59,11 @@ func TestPostMessage_ServeHTTP(t *testing.T) {
 				OnFailedPollWaitSecs: 1,
 			}
 
+			err := a.initClient()
```
evankanderson: Slight preference for the 1-line format:

```golang
if err := a.initClient(); err != nil {
    t.Errorf("....")
}
```

Note that this changes the scope off `err` to be only within the `if` block, which should be harmless here, but does sometimes mean you can't use this format.

pkg/adapter/cronjobevents/adapter.go
```diff
@@ -49,6 +49,22 @@ type Adapter struct {
 	client client.Client
 }
 
+// Initialize cloudevent client
+func (a *Adapter) initClient() error {
+	if a.client == nil {
+		var err error
+		if a.client, err = client.NewHTTPClient(
```
evankanderson: In this case, I think it would be more fluent to do the opposite:

```golang
a.client, err := client.NewHTTPClient(
      ....
      ....)
if err != nil {
    return err
}
```

Two rubrics for choosing between the two forms:

1. Avoid the inline `if` assignment if the function call is multiple lines; it's harder to follow the `if` block that way.
2. Which is fewer lines of code? In this case, needing to pre-declare `var err error` in order to be able to assign to `a.client` removes all the benefit of inlining the assignment in the `if` statement.

pkg/adapter/github/adapter.go
```diff
@@ -56,15 +59,20 @@ func (ra *Adapter) HandleEvent(payload interface{}, header webhooks.Header) {
...
+				client.WithTarget(ra.Sink),
+				client.WithHTTPBinaryEncoding(),
+				client.WithUUIDs(),
+				client.WithTimeNow(),
+			); err != nil {
+				log.Printf("failed to create cloudevent client: %s", err)
+			}
+		})
+	}
+	if ra.client == nil {
```
evankanderson: This double-check on `ra.client` is a bit strange. (I realize that the current structure requires this with `sync.Once`. Maybe extract this into a function, or call the `sync.Once` unconditionally?)

pkg/adapter/github/adapter.go
```diff
@@ -56,15 +59,20 @@ func (ra *Adapter) HandleEvent(payload interface{}, header webhooks.Header) {
 
 func (ra *Adapter) handleEvent(payload interface{}, hdr http.Header) error {
 	if ra.client == nil {
```
grantr: nit: no need for this outer if.

```go
func (ra *Adapter) handleEvent(payload interface{}, hdr http.Header) error {
  ra.initClientOnce.Do(func() {
    // ...
  })
  if ra.client == nil {
  // ...
```

grantr:
/hold
for @evankanderson's comments

pkg/adapter/github/adapter.go
grantr: I think calling `once.Do()` every time is the right answer. https://github.com/knative/eventing-sources/pull/240#discussion_r264792991

pkg/adapter/cronjobevents/adapter.go
whynowy: @evankanderson - current golang doesn't support `a.client, err := client.NewHTTPClient....`

see the issue: 
https://github.com/golang/go/issues/30318

pkg/adapter/github/adapter.go
whynowy: That is something I'm not very comfortable to use `Sync.once`, it can not do error handling and retry well. If it fails, `ra.client` will always be `nil` without any chance to retry. Anyways, I removed the check based on your comments.

grantr: That seems like a good reason to have an initialization method (like the `Start` method the other adapters have) so the caller can receive an error when the client is unable to start.

pkg/adapter/cronjobevents/adapter.go
whynowy: and this:  https://github.com/golang/go/issues/6842

pkg/adapter/github/adapter.go
whynowy: In this case we need to make the initialization method public, and call it in `main.go` like below, it doesn't seem like very good. Is there any other way to do it?

```
	ra := &github.Adapter{
		Sink: *sink,
	}

        # Init cloud event client
        InitCloudEventClient()

	hook := gh.New(&gh.Config{Secret: secretToken})
	hook.RegisterEvents(ra.HandleEvent,
		gh.CommitCommentEvent,
```

grantr: Could all this be wrapped up in a blocking `Start`? But this seems like something to open an issue about rather than blocking the PR on it.

pkg/adapter/github/adapter.go
```diff
@@ -71,9 +71,6 @@ func (ra *Adapter) handleEvent(payload interface{}, hdr http.Header) error {
 			}
 		})
 	}
-	if ra.client == nil {
-		return fmt.Errorf("failed to create cloudevent client")
-	}
```
grantr: I think I was expecting this second `if ra.client == nil` to remain, but remove the first `if ra.client == nil` around `ra.initClientOnce.Do()`.

grantr: Is that what you expected @evankanderson?

whynowy: 👍 Makes sense, It's better! I'll update it first :)

PR #243, opened 2019-03-11
Enable specifying https github webhooks

dicarlo2:
## Proposed Changes

  * Add spec property `secure` which when set to true causes the webhook to be configured using https rather than the default http.

**Release Note**

```release-note
Configure https GitHub webhooks using the `secure` property.
```

### DIFF
--- a/pkg/reconciler/githubsource/githubsource_test.go
+++ b/pkg/reconciler/githubsource/githubsource_test.go
@@ -218,6 +218,60 @@ var testCases = []controllertesting.TestCase{
 		},
 		IgnoreTimes: true,
 	}, {
+		Name:       "valid githubsource, secure repo webhook created",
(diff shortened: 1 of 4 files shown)


n3wscott:
nice.

/ok-to-test

pkg/reconciler/githubsource/githubsource_test.go
```diff
@@ -794,6 +852,10 @@ func (client mockWebhookClient) Delete(ctx context.Context, options *webhookOpti
 		return fmt.Errorf(`expected webhook repository of "%s", got "%s"`,
 			data.expectedRepo, options.repo)
 	}
+	if data.expectedSecure != options.secure {
+		return fmt.Errorf(`expected webhook secure of "%v", got "%v"`,
```
vaikas-google: total nit, but since these are booleans, maybe %t here?

PR #248, opened 2019-03-14
Use cloudevents sdk 0.4.0

n3wscott:
Dep on `knative/pkg/cloudevents` is gone. 🎉 

**Release Note**

```release-note
All sources now emit version 0.2 of CloudEvents.
```


pkg/adapter/kubernetesevents/adapter.go
```diff
@@ -98,16 +101,16 @@ func (a *Adapter) addEvent(new interface{}) {
 	logger := logging.FromContext(context.TODO()).With(zap.Any("eventID", event.ObjectMeta.UID), zap.Any("sink", a.SinkURI))
 	logger.Debug("GOT EVENT", event)
 
+	var err error
+	a.initClientOnce.Do(func() {
```
vaikas-google: Just curious about doing this instead of creating this in Start() method like where you create the kubeClient.

contrib/gcppubsub/pkg/adapter/adapter.go
```diff
@@ -89,40 +94,34 @@ func (a *Adapter) receiveMessage(ctx context.Context, m PubSubMessage) {
 	}
 }
 func (a *Adapter) postMessage(ctx context.Context, logger *zap.SugaredLogger, m PubSubMessage) error {
+	var err error
+	a.initClientOnce.Do(func() {
```
vaikas-google: why not do this in Start? I think it would be better to fail during start. But guess the fact that I'm asking this question here and in k8sevents means that there should probably be some comments added.

n3wscott: whyyyy not. ok done.

pkg/adapter/github/adapter.go
```diff
@@ -42,35 +40,28 @@ const (
...
-			client.WithTarget(ra.Sink),
-			client.WithHTTPBinaryEncoding(),
-			client.WithUUIDs(),
-			client.WithTimeNow(),
-		); err != nil {
-			log.Printf("failed to create cloudevent client: %s", err)
-		}
+func (a *Adapter) handleEvent(payload interface{}, hdr http.Header) error {
+	var err error
+	a.initClientOnce.Do(func() {
```
vaikas-google: this too please :)

n3wscott: this one does not have a start method. 😕

vaikas-google: I think it would be better if this file would have a NewAdapter method and we'd do it there. But fine...

PR #250, opened 2019-03-14
Use New func to protect race condition for cloudevent client.

n3wscott:
Followup to #248 

**Release Note**

```release-note
NONE
```

/assign @vaikas-google


cmd/github_receive_adapter/main.go
```diff
@@ -56,8 +56,9 @@ func main() {
 
 	log.Printf("Sink is: %q", *sink)
 
-	ra := &github.Adapter{
-		SinkURI: *sink,
+	ra, err := github.New(*sink)
+	if err != nil {
+		log.Fatalf("Failed to create github adaptor: %s", err.Error())
```
evankanderson: I think you want "adapter": https://writingexplained.org/adapter-vs-adaptor-difference

("Adapter" is overwhelmingly more popular in American English; British English is more evenly split, but there are no specific rules differentiating the two.)

n3wscott: every time...

n3wscott: fixed.

PR #285, opened 2019-03-25
wire labels and annotations through to the deployment

vaikas-google:
## Proposed Changes

  * If labels or annotations are specified on the containersource, propagate them to the podspecs. This allows us to do things like control Istio outbound ip ranges.

**Release Note**

```release-note
If containersource has labels or annotations, propagate them to the podspecs of the reified containersource deployments.
Change label denoting source from: `source` to `eventing.knative.dev/source`
This label can not be overwritten, others will be.
```

### DIFF
--- a/pkg/reconciler/containersource/containersource.go
+++ b/pkg/reconciler/containersource/containersource.go
@@ -96,16 +96,14 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) error
 
 	source.Status.InitializeConditions()
 
-	// These are the annotations / labels we always apply
-	labels := map[string]string{"source": source.Name}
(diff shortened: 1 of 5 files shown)


pkg/reconciler/containersource/containersource.go
```diff
@@ -96,13 +96,31 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) error
 
 	source.Status.InitializeConditions()
 
+	// These are the annotations / labels we always apply
+	labels := map[string]string{"source": source.Name}
```
n3wscott: We have been using qualified names, I would suggest:

```
labels := map[string]string{"eventing.knative.dev/source": source.Name}
```

n3wscott:
/lgtm

nice work.

PR #298, opened 2019-03-27
Clean up args handling, remove todo

iancoffey:
Address webhook related TODO by cleaning up the args as suggested by the comments.

https://github.com/knative/eventing-sources/blob/master/pkg/reconciler/githubsource/githubsource.go#L203-L213
https://github.com/knative/eventing-sources/blob/master/pkg/reconciler/githubsource/githubsource.go#L240-L250

## Proposed Changes

* Two small args related changes to remove inline TODO regarding sloppy arguments.
**Release Note**

```release-note

NONE
```

### DIFF
--- a/pkg/reconciler/githubsource/githubsource.go
+++ b/pkg/reconciler/githubsource/githubsource.go
@@ -83,6 +83,15 @@ type reconciler struct {
 	webhookClient       webhookClient
 }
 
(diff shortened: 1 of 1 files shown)


vaikas-google:
Thanks for doing this!
/lgtm
/approve

PR #300, opened 2019-03-27
Event Registry - Sources Part 1

nachocano:
First part of sources for the Registry feature request: https://github.com/knative/eventing/issues/929

## Important
  * This still doesn't address the changes of CloudEvents source and subject attributes. There is still some debate of what should be put in each. This PR just sets the framework necessary to easily set those attributes in a subsequent PR.
  * Changes to Sources CRD to include description and schema attributes are not there yet.

## Proposed Changes
  * Creating EventTypes in the following source controllers, from which we know (at least some of) the events they can produce:
     * GitHub
     * AWS SQS
     * GCPPubSub
     * Kafka
  * Only creating the EventTypes for Broker sinks. 
 * Created a common EventType "reconciler" helper to avoid duplicating code across the source controllers.
  * Adding dependency of the eventing repo  (and updating pkg)
 

**Release Note**

```release-note
On installing a source CO (GitHub, AWS SQS, GCPPubSub, or Kafka), its EventTypes will be added to the Registry.
```


nachocano:
/cc @vaikas-google @grantr @Harwayne @akashrv @n3wscott @duglin 

Here is the sources part of the Registry stuff I've been experimenting with. Comments and feedback are highly appreciated.

Gopkg.toml
```diff
@@ -103,3 +107,9 @@ required = [
 [[constraint]]
   name = "github.com/apache/camel-k"
   version = "0.2.0"
+
+# Dependency on Registry
```
nachocano: This will change of course, just pointing to my branch

Gopkg.toml
```diff
@@ -103,3 +107,9 @@ required = [
...
+[[constraint]]
+  name = "github.com/knative/eventing"
+  source = "github.com/nachocano/eventing"
```
nachocano: this will change, just pointing to my branch to showcase stuff

nachocano:
@Abd4llA this is the sources part of the Registry proposal. Also WIP, but it'd be great to have your input.
Thanks!

duglin:
> /cc @vaikas-google @grantr @Harwayne @akashrv @n3wscott @duglin
> 
> Here is the sources part of the Registry stuff I've been experimenting with. Comments and feedback are highly appreciated.

Which file has the github types? I couldn't find it when I searched

nachocano:
> Which file has the github types? I couldn't find it when I searched

You mean this one? https://github.com/knative/eventing-sources/blob/a81e4bf176fdde1549c7a6e339b3baf185e7e1b6/pkg/apis/sources/v1alpha1/githubsource_types.go

I'm not changing any of the type CRDs as of now. Just using them as is for an initial cut.
With the GitHubSpec fields I'm creating the EventTypes CRs in the GitHub source controller.

nachocano:
Problem here: https://github.com/knative/test-infra/issues/638

/retest

nachocano:
@vaikas-google @grantr @Harwayne @n3wscott 
This is ready to start reviewing. Please look at the PR description, this still doesn't include the change to source and subject, as it seems we are still not in full agreement.

nachocano:
The problem seems to be that the e2e script pulls the eventing nightly build:

readonly KNATIVE_BASE_YAML_SOURCE=https://storage.googleapis.com/knative-nightly/@/latest
KNATIVE_EVENTING_RELEASE=${KNATIVE_BASE_YAML_SOURCE/@/eventing}/release.yaml

As we've just merged the registry change in eventing, it will be in the next nightly build, and therefore tests should start passing (hopefully)
Will wait until tomorrow and rerun them.

nachocano:
@vaikas-google @Harwayne @grantr @n3wscott friendly call for reviews :)

contrib/gcppubsub/pkg/reconciler/gcppubsubsource_test.go
```diff
@@ -340,6 +384,35 @@ func getSource() *sourcesv1alpha1.GcpPubSubSource {
 	return obj
 }
 
+func getEventType() *eventingv1alpha1.EventType {
```
n3wscott: this should be a method called `NewEventType` in the `./resources` directory for this reconciler.

n3wscott: Oh, I am sorry, this is in the test.

IGNORE!

pkg/reconciler/eventtype/eventtype.go
```diff
@@ -0,0 +1,208 @@
...
+		if err := controllerutil.SetControllerReference(owner, &eventType, r.Scheme); err != nil {
+			return nil, err
+		}
+		eventTypes = append(eventTypes, eventType)
+	}
+	return eventTypes, nil
+}
+
+// makeEventType creates the in-memory representation of the EventType.
+func (r *Reconciler) makeEventType(arg *EventTypeArgs, namespace string, lbl map[string]string) eventingv1alpha1.EventType {
```
n3wscott: _This_ should be in the resource dir for this reconciler.

pkg/reconciler/eventtype/eventtype.go
```diff
@@ -0,0 +1,208 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// Only allow alphanumeric, '-' or '.'.
+var validChars = regexp.MustCompile(`[^-\.a-z0-9]+`)
+
+// Reconciler reconciles EventTypes of different sources.
+type Reconciler struct {
```
n3wscott: Is this something that can be pulled in from eventing?

pkg/reconciler/githubsource/githubsource.go
```diff
@@ -58,17 +61,23 @@ func Add(mgr manager.Manager, logger *zap.SugaredLogger) error {
...
+		receiveAdapterImage: receiveAdapterImage,
+		webhookClient:       gitHubWebhookClient{},
+		eventTypeReconciler: eventtype.Reconciler{
+			Scheme: scheme,
+		},
+	}
 
+	log.Println("Adding the GitHub Source controller.")
 	p := &sdk.Provider{
-		AgentName: controllerAgentName,
```
n3wscott: I would fold this all back up again and just call mgr.GetScheme() twice.

pkg/reconciler/githubsource/githubsource.go
```diff
@@ -183,7 +193,20 @@ func (r *reconciler) reconcile(ctx context.Context, source *sourcesv1alpha1.GitH
 			}
 			source.Status.WebhookIDKey = hookID
 		}
+	} else {
+		return nil
 	}
+
+	// Only create EventTypes for Broker sinks.
+	if source.Spec.Sink.Kind == "Broker" {
```
n3wscott: why?

pkg/reconciler/githubsource/githubsource.go
```diff
@@ -327,6 +350,38 @@ func (r *reconciler) getOwnedService(ctx context.Context, source *sourcesv1alpha
...
+		args = append(args, arg)
+	}
+	return &eventtype.ReconcilerArgs{
+		EventTypes: args,
+		Namespace:  source.Namespace,
+		Labels:     getLabels(source),
+	}
+}
+
+func getLabels(src *sourcesv1alpha1.GitHubSource) map[string]string {
```
n3wscott: this should be in the resource dir too.

pkg/reconciler/eventtype/eventtype.go
nachocano: I may be able to put it there, as we might need it for the core sources as well. 
Do you want me to do that? 
The downside with that is that we will have a dependency to eventing of other things than just API types. 
@grantr and I were thinking of pushing for an api repo, and only depend on that in the future. I guess that will take some time to get approved, so we might be able to move this there.

pkg/reconciler/eventtype/eventtype.go
nachocano: this is not technically a reconciler but rather a helper not to duplicate code across sources controllers.
But yes, I can move it there...

pkg/reconciler/githubsource/githubsource.go
nachocano: This is what we said in the proposal. Registry is mainly for Broker/Trigger model, at least for the MVP. If you are just sending stuff to another sink, then the EventType won't be added to the registry. 

There may be a more "typed" way to do this check?

pkg/reconciler/githubsource/githubsource.go
nachocano: will do

nachocano: Done

pkg/reconciler/githubsource/githubsource.go
nachocano: Done, rollbacked for all four source controllers that I changed.

pkg/reconciler/githubsource/githubsource.go
n3wscott: Why filter?

What if the source goes `source -> channel -> broker`?

pkg/reconciler/eventtype/eventtype.go
nachocano: Actually, do you think this file should go inside pkg/reconciler/eventtype? 
Or just leave it in pkg/reconciler? as a sort of helper for all other reconcilers, as this is technically not a reconciler.
Or maybe just put it outside, pkg/eventtype?

contrib/gcppubsub/pkg/reconciler/gcppubsubsource.go
```diff
@@ -165,6 +172,17 @@ func (r *reconciler) Reconcile(ctx context.Context, object runtime.Object) error
...
+	// Only create EventTypes for Broker sinks.
+	if src.Spec.Sink.Kind == "Broker" {
+		err = r.reconcileEventTypes(ctx, src)
+		if err != nil {
+			logger.Error("Unable to reconcile the event types", zap.Error(err))
+			return err
+		}
+	}
+	// We mark EventTypes in order to have the source Ready, even though the Sink might haven't been a Broker.
+	src.Status.MarkEventTypes()
```
grantr: I recommend removing `EventTypesProvided` from the set of conditions that indicate Ready. It should still be a condition, but it shouldn't be part of source readiness.

The registry is optional. The code here acknowledges this by setting `EventTypesProvided` true when the sink is not a Broker, but Brokers may have an accept everything policy, which should seemingly produce the same behavior.

I think it's clearer for the user to define source readiness as "capable of sending events". If the sink is rejecting events, that's a problem (and possibly an error condition), but the source is still ready.

pkg/reconciler/githubsource/githubsource.go
nachocano: Our current EventType CRD has a mandatory `broker` field. 
I guess we should have called it `sink` if we were aiming for other cases.
I'm somewhat reluctant to change it right now, as that is what the community accepted. 
But I'm open to discussions. What do other people think @vaikas-google @grantr @Harwayne

contrib/gcppubsub/pkg/reconciler/gcppubsubsource.go
```diff
@@ -281,6 +299,29 @@ func (r *reconciler) deleteSubscription(ctx context.Context, src *v1alpha1.GcpPu
...
+func (r *reconciler) reconcileEventTypes(ctx context.Context, src *v1alpha1.GcpPubSubSource) error {
+	args := r.newEventTypesReconcilerArgs(src)
+	return r.eventTypeReconciler.ReconcileEventTypes(ctx, src, args)
+}
+
+func (r *reconciler) newEventTypesReconcilerArgs(src *v1alpha1.GcpPubSubSource) *eventtype.ReconcilerArgs {
+	arg := &eventtype.EventTypeArgs{
+		Type: v1alpha1.GcpPubSubSourceEventType,
+		// Using the google cloud project and the topic as source.
+		// This should match what is populated in the adapter.
```
grantr: If there were a function in the gcppubsub types package producing this string, we wouldn't need this coupling.

cmd/github_receive_adapter/main.go
```diff
@@ -33,6 +33,9 @@ const (
 
 	// Environment variable containing GitHub secret token
 	envSecret = "GITHUB_SECRET_TOKEN"
+
+	// Environment variable containing information about the origin of the event
+	envOwnerRepo = "GITHUB_OWNER_REPO"
```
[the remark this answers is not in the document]
nachocano: I'm setting it in pkg/reconciler/githubsource/resources/service.go. 
This is actually more for the follow up PR on properly setting source and subject, I guess I missed removing it when I did the split. But it doesn't hurt leaving it here anyway.
If you want I can remove it.

cmd/manager/main.go
```diff
@@ -20,6 +20,7 @@ import (
 	"log"
 
 	"github.com/knative/eventing-sources/pkg/apis"
+	_ "github.com/knative/eventing-sources/pkg/apis/eventing/v1alpha1"
```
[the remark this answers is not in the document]
nachocano: The one below was already there. I added the new one as we are now depending on eventing and we need the eventing API types to be added to the scheme (or that is my understanding). I think that is done by some magic in pkg/apis. If you don't have that, then controllers cannot find Service, EventType, etc.

contrib/gcppubsub/pkg/apis/sources/v1alpha1/gcp_pubsub_types.go
```diff
@@ -74,6 +74,14 @@ type GcpPubSubSourceSpec struct {
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
 }
 
+const (
+	// GcpPubSubSourceEventType is the GcpPubSub CloudEvent type, in case PubSub doesn't send a
+	// CloudEvent itself.
+	GcpPubSubSourceEventType = "google.pubsub.topic.publish"
+	// GcpPubSubSourceEventSourceTemplate is the GcpPubSub CloudEvent source template.
+	GcpPubSubSourceEventSourceFormat = "//pubsub.googleapis.com/%s/topics/%s"
```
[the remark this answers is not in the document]
nachocano: Makes sense! Done, thanks!

contrib/gcppubsub/pkg/reconciler/gcppubsubsource.go
nachocano: Makes total sense! Thanks for the comment. Will do that from the other controllers I changed as well.

contrib/gcppubsub/pkg/reconciler/gcppubsubsource.go
nachocano: Done!

pkg/reconciler/eventtype/eventtype.go
nachocano: Actually, I think it would be better to have this stuff here (and possibly do @grantr changes). 
The only source from eventing that might need to create EventTypes is CronJob, and it can do it as we typically do when creating resources from controllers. If we later on see a need for this in eventing, I can move it there.  
If there is a strong objection, we can further discuss. I'll mark this as resolve for now

cmd/github_receive_adapter/main.go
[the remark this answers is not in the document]
nachocano: I'll mark it as resolved for now

config/201-clusterrole.yaml
```diff
@@ -87,3 +87,10 @@ rules:
   resources:
   - events
   verbs: *everything
+
+# EventTypes admin
+- apiGroups:
```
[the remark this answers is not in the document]
vaikas-google: sorry, I thought I had already made this review comment a while ago, but must have failed to hit the submit button :(

pkg/reconciler/eventtype/eventtype.go
```diff
@@ -0,0 +1,208 @@
...
+		if err := r.Client.List(ctx, opts, el); err != nil {
+			return nil, err
+		}
+
+		for _, e := range el.Items {
+			if metav1.IsControlledBy(&e, owner) {
+				eventTypes = append(eventTypes, e)
+			}
+		}
+		if el.Continue != "" {
```
grantr: I think we established elsewhere that if we don't set a `limit` option we don't have to do pagination, so I think this pagination code can be removed. cc @Harwayne

pkg/reconciler/githubsource/githubsource.go
vaikas-google: I'd prefer to leave this as is for now. I could see a use case where you could list event types that are wired for a given channel though. That's what the agreed upon design was so we can revisit at a later date.

config/201-clusterrole.yaml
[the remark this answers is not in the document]
vaikas-google: And this can be done as a follow on, but just a thought.

pkg/reconciler/eventtype/eventtype.go
grantr: IMO this whole thing is effectively its own reconciler, and should have its own `/reconciler` directory with a `/resources` subdir. But again I don't know how elegant that will be in practice.

config/201-clusterrole.yaml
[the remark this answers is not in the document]
nachocano: was just fixing merge conflicts, sorry.
Yes, let me talk with Adam so that he explains me what are the aggregated rules, and I can certainly do it as a follow up PR...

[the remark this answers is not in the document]
nachocano: thanks for the comment!

contrib/gcppubsub/pkg/apis/sources/v1alpha1/gcp_pubsub_types.go
```diff
@@ -74,6 +76,17 @@ type GcpPubSubSourceSpec struct {
...
 }
 
+const (
+	// GcpPubSubSourceEventType is the GcpPubSub CloudEvent type, in case PubSub doesn't send a
+	// CloudEvent itself.
+	GcpPubSubSourceEventType = "google.pubsub.topic.publish"
+)
+
+// GetGcpPubSubSource returns the GcpPubSub CloudEvent source value.
+func GetGcpPubSubSource(googleCloudProject, topic string) string {
```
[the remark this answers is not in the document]
nachocano: I used the Get because GcpPubSubSource gives conflicts as the struct is called like that. 
Maybe GcpPubSubCloudEventSource ?

pkg/reconciler/githubsource/githubsource.go
nachocano: Agree. Will leave it as is for now, and then we can iterate over it.

pkg/reconciler/eventtype.go
```diff
@@ -0,0 +1,193 @@
...
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+	}
+	return nil
+}
+
+// getEventTypes returns the EventTypes controlled by 'owner'.
+func (r *EventTypeReconciler) getEventTypes(ctx context.Context, namespace string, lbs map[string]string, owner metav1.Object) ([]eventingv1alpha1.EventType, error) {
+	eventTypes := make([]eventingv1alpha1.EventType, 0)
```
grantr: I think we normally do `eventTypes := []eventingv1alpha1.EventType{}`

pkg/reconciler/eventtype.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// getSyncEventTypes computes the EventTypes that need to be created and/or deleted based on the difference between
+// 'expected' and 'current'. It does so using all the EventType.Spec fields but Description.
+func (r *EventTypeReconciler) getEventTypesToSync(current []eventingv1alpha1.EventType, expected []eventingv1alpha1.EventType) *syncEventTypes {
+	eventTypesToSync := &syncEventTypes{
+		ToCreate: make([]eventingv1alpha1.EventType, 0),
+		ToDelete: make([]eventingv1alpha1.EventType, 0),
+	}
+	keyFunc := func(eventType *eventingv1alpha1.EventType) string {
```
grantr: What do you think about making this a package function instead? Then it's clear that it doesn't rely on method-local scope. I don't have a strong preference either way.

pkg/reconciler/eventtype/eventtype.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Reconciler reconciles EventTypes of different sources.
+type Reconciler struct {
+	Client client.Client
+	Scheme *runtime.Scheme
+}
+
+// ReconcilerArgs are the arguments needed to reconcile EventTypes.
+type ReconcilerArgs struct {
+	EventTypes []*EventTypeArgs
```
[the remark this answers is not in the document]
nachocano: Made that change, but removed the EventTypeArgs struct, as it is not needed. Is just the spec.

contrib/kafka/pkg/reconciler/kafkasource.go
```diff
@@ -164,6 +179,29 @@ func (r *reconciler) getReceiveAdapter(ctx context.Context, src *v1alpha1.KafkaS
 	return nil, apierrors.NewNotFound(schema.GroupResource{}, "")
 }
 
+func (r *reconciler) reconcileEventTypes(ctx context.Context, src *v1alpha1.KafkaSource) error {
+	args := r.newEventTypesReconcilerArgs(src)
+	return r.eventTypeReconciler.ReconcileEventTypes(ctx, src, args)
+}
+
+func (r *reconciler) newEventTypesReconcilerArgs(src *v1alpha1.KafkaSource) *eventtype.ReconcilerArgs {
+	args := make([]*eventtype.EventTypeArgs, 0)
```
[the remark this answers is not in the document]
nachocano: This changed to specs := make([]eventingv1alpha1.EventTypeSpec, 0)... 
If I just do:
`[]eventingv1alpha1.EventTypeSpec{}`
Goland gives a warning: Empty slice declaration via literal

pkg/reconciler/eventtype.go
```diff
@@ -0,0 +1,193 @@
...
+	for _, e := range expected {
+		if et, ok := currentMap[keyFunc(&e)]; !ok {
+			// If it's not in the currentMap, we need to create it.
+			eventTypesToSync.ToCreate = append(eventTypesToSync.ToCreate, e)
+		} else {
+			// If it's in the currentMap and there are more than one,
+			// then we just keep the first one, and delete the others.
+			if len(et) > 1 {
+				for i := 1; i < len(et); i++ {
+					eventTypesToSync.ToDelete = append(eventTypesToSync.ToDelete, et[i])
```
grantr: I wonder if there are race conditions here. I haven't thought it through carefully, but if two controllers are trying to delete event types, is it possible they could choose to keep different ones, thus deleting all of them? This list only includes types that are owned by a single source object, right?

I guess if we were worried about this, one possible workaround is to sort by creation timestamp (plus some deterministic secondary key like UID to break ties) so that the oldest one is always kept. Seems like a lot of work though.

@nachocano if the "clean up duplicates" task were done by a different controller whose only job was cleaning up EventType dupes, would this function be simpler (or be necessary at all)?

contrib/kafka/pkg/reconciler/kafkasource.go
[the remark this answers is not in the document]
nachocano: Shall I change it anyway?

pkg/reconciler/eventtype.go
```diff
@@ -0,0 +1,193 @@
...
+	return eventTypesAsMap
+}
+
+// Converts 'name' to a valid DNS1123 subdomain, required for object names in K8s.
+func ToDNS1123Subdomain(name string) string {
+	// If it is not a valid DNS1123 subdomain, make it a valid one.
+	if msgs := validation.IsDNS1123Subdomain(name); len(msgs) != 0 {
+		// If the length exceeds the max, cut it and leave some room for a potential generated UUID.
+		if len(name) > validation.DNS1123SubdomainMaxLength {
+			name = name[:validation.DNS1123SubdomainMaxLength-10]
```
grantr: Should this `10` be a constant? Maybe easier to document.

contrib/gcppubsub/pkg/reconciler/gcppubsubsource.go
```diff
@@ -66,16 +64,22 @@ func Add(mgr manager.Manager, logger *zap.SugaredLogger) error {
 		return fmt.Errorf("required environment variable '%s' not defined", raImageEnvVar)
 	}
 
+	scheme := mgr.GetScheme()
+	r := &reconciler{
+		scheme:              scheme,
+		pubSubClientCreator: gcpPubSubClientCreator,
+		receiveAdapterImage: raImage,
+		eventTypeReconciler: eventtype.Reconciler{
```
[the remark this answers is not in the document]
nachocano: Talking with Grant this seems like a great idea to reduce boilerplate code from the current controllers. We agree we can do this in a follow up.

contrib/kafka/pkg/reconciler/kafkasource.go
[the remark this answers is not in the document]
grantr: Nope, if the IDE complains it's probably right :)

contrib/gcppubsub/pkg/apis/sources/v1alpha1/gcp_pubsub_types.go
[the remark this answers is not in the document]
grantr: Oh I didn't realize there would be a conflict. Keep it as is then.

pkg/reconciler/eventtype.go
grantr: (You can ignore this comment given https://github.com/knative/eventing-sources/pull/300#discussion_r280204402)

pkg/reconciler/eventtype/eventtype.go
nachocano: Will address that structure change in the follow up once we go with your approach to reduce boilerplate code.

contrib/gcppubsub/pkg/apis/sources/v1alpha1/gcp_pubsub_types.go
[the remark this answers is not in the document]
nachocano: will do, thanks!

contrib/kafka/pkg/reconciler/kafkasource.go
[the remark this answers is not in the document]
nachocano: OK, will leave it as is then

pkg/reconciler/eventtype/eventtype.go
nachocano: Removed, wasn't aware of that. Just copy/pasting from other places

pkg/reconciler/eventtype.go
nachocano: Done, made it a package function instead! Thanks for the comment

pkg/reconciler/eventtype.go
nachocano: Done! thanks!

pkg/reconciler/eventtype.go
nachocano: I think this deletion logic is not needed anymore. As of now, EventType spec fields that determine their uniqueness (type,source,schema,broker) are immutable. The only thing that could happen is if someone deletes them by mistake, if so, then we recreate them. 
Will remove this deletion logic and simplify this function.

> This list only includes types that are owned by a single source object, right?

Yes, that is correct.

pkg/apis/eventing/v1alpha1/addtoscheme_eventing_v1alpha1.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "github.com/knative/eventing-sources/pkg/apis"
+
+import (
+	eventingv1alpha1 "github.com/knative/eventing/pkg/apis/eventing/v1alpha1"
+)
+
+func init() {
+	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
+	apis.AddToSchemes = append(apis.AddToSchemes, eventingv1alpha1.SchemeBuilder.AddToScheme)
```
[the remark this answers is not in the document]
nachocano: I'm not sure what is the right way of doing this. You might be right and we should add them in cmd. In tests we are doing so. 
@n3wscott ?

config/201-clusterrole.yaml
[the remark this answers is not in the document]
nachocano: Created an issue to track this https://github.com/knative/eventing-sources/issues/389

nachocano:
> Got through everything!

Thanks a lot @grantr . Your review was very helpful! Will let you know when is ready for a second pass

config/201-clusterrole.yaml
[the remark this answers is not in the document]
Harwayne: I'm not sure that an aggregated role makes sense here, at least not just for EventTypes. Normally we use aggregated roles to say things like 'verb all nouns', where verb is read or write, and nouns are multiple object types, such as 'sources', 'addressables', etc. 

As this change adds EventTypes, and that is a single, known CRD, I don't think it would benefit from an aggregated role.

pkg/apis/eventing/v1alpha1/addtoscheme_eventing_v1alpha1.go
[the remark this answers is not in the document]
n3wscott: I think the way I am doing it is correct, it is how we get serving schemes.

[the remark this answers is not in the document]
grantr: How does serving get the pkg types into its scheme?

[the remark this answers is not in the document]
nachocano: Hmmm, I see a different [thing](https://github.com/knative/serving/blob/ede8da2a1a0c2e43d645dc8ad16e934d1fa3c4c4/pkg/apis/serving/v1alpha1/register.go#L46).

@grantr do you think we maybe track this as a separate issue to be solved in a follow up? Same as the better eventType reconciler struct?

nachocano:
Tested gcppubsub, kafka, and github... Seems to be working fine.
Cannot easily test aws sqs.

```
NAME                                          TYPE                                     SOURCE                                                          SCHEMA   BROKER    DESCRIPTION   READY   REASON
dev.knative.kafka.event-cjvcr                 dev.knative.kafka.event                  topic2                                                                   default                 True    
dev.knative.kafka.event-tdt48                 dev.knative.kafka.event                  knative-demo-topic                                                       default                 True    
dev.knative.source.github.pullrequest-pkb8c   dev.knative.source.github.pull_request   nachocano/eventing                                                       default                 True    
dev.knative.source.github.push-vmbl2          dev.knative.source.github.push           nachocano/eventing                                                       default                 True    
google.pubsub.topic.publish-hrxhh             google.pubsub.topic.publish              //pubsub.googleapis.com/knative-project-228222/topics/testing            default                 True    
```

pkg/apis/eventing/v1alpha1/addtoscheme_eventing_v1alpha1.go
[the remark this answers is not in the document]
nachocano: Created an issue https://github.com/knative/eventing-sources/issues/390 to separately track this.

nachocano:
@grantr @vaikas-google @n3wscott  
this is ready for a second pass. There are a couple of separate follow up issues that I can take care later on, but the basic functionality herein seems to be working (see `kubectl get eventtypes` output above). The main one is a refactoring of the eventtype reconciler proposed by Grant, which makes a lot of sense. I guess I didn't get it right away. 

The next PR should contain the changes to source and subject.

nachocano:
Made a small refactor to make it easier for the bigger one later on. Added the eventtype package again, inside reconciler, with its own resources dir.

nachocano:
@grantr @vaikas-google @n3wscott friendly ping to lgtm (or keep on adding comments) so that I can continue with the follow ups.

vaikas-google:
I already approved, sorry. should've said explicitly, letting @grantr @n3wscott take a look since they had some suggestiongs.

pkg/apis/eventing/v1alpha1/addtoscheme_eventing_v1alpha1.go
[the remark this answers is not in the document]
grantr: > Hmmm, I see a different thing.

`addKnownTypes` is the internal method called by SchemeBuilder's `AddToScheme`. The whole thing is pretty confusing.

pkg/reconciler/eventtype/eventtype.go
grantr: :+1: Please also remove these 3 lines starting at https://github.com/knative/eventing-sources/pull/300/commits/dca6574b0ee80007e49633d0756a458c2ca0b9bf#diff-664142aa17d510b591fb692036e3bf87R97:

```go
// Set Raw because if we need to get more than one page, then we will put the continue token
// into opts.Raw.Continue.
Raw: &metav1.ListOptions{},
```

nachocano: Done, thanks!

nachocano:
> Just the one change left from me (remove `Raw`) then

Done!

PR #308, opened 2019-03-28
KafkaSource does NO LONGERignore failures when POSTing to sinks

### LINKED ISSUE #304: Sources should not ignore failures when POSTing to sinks
## Expected Behavior
A source's receive adapter should not ack (whatever form this takes - commit offsets, responding with HTTP 200, etc) receiving of an event unless it successfully POSTs that event to the Sink.

## Actual Behavior
Several of the sources ack receipt of the event and then subsequently log a message but otherwise drop events entirely if POSTing of the event to the sink fails.

  1. Deploy the Kafka or GitHub event source, for example, and have it sink to a service that always returns an error
  2. Send an event and you'll see the offset get committed in Kafka or the webhook delivery succeed in GitHub even though we ended up not delivering the event successfully.

## Additional Info

GCPPubSub and AWS SQS sources appear to do the right thing! Someone more intimately familiar with those sources may want to confirm, but neither of them ack if POSTing to the sink fails.

Now, for the ones that do the wrong thing:

### Kafka
https://github.com/knative/eventing-sources/blob/8ff3a26a58debffb0225bb7afeb932d12f27115a/contrib/kafka/pkg/adapter/adapter.go#L82 and https://github.com/knative/eventing-sources/blob/8ff3a26a58debffb0225bb7afeb932d12f27115a/contrib/kafka/pkg/adapter/adapter.go#L162-L164

We should not commit offsets for failures.

### GitHub
https://github.com/knative/eventing-sources/blob/8ff3a26a58debffb0225bb7afeb932d12f27115a/pkg/adapter/github/adapter.go#L60-L62

We should respond with an error code instead of HTTP 20x in the GitHub webhook so that GitHub at least knows delivery failed and can apply any server-side retry policy they have.

### Cron Job
https://github.com/knative/eventing-sources/blob/8ff3a26a58debffb0225bb7afeb932d12f27115a/pkg/adapter/cronjobevents/adapter.go#L98-L100

Not entirely sure what this one should do, but adding it for completeness sake.

### Kubernetes Events
https://github.com/knative/eventing-sources/blob/8ff3a26a58debffb0225bb7afeb932d12f27115a/pkg/adapter/kubernetesevents/adapter.go#L112-L114

We can probably do something similar to our controllers where a failure puts the item back in the queue to be reconciled? I realize we're not doing a full reconcile loop here, but the same principle perhaps applies? I don't know enough about direct Informer usage to do anything beyond speculate.

### Camel
It's not clear. It's worth reviewing what it does, but it's a bit different from all these others.

matzew:
Fixes #304 (Kafka)

## Proposed Changes

  * the `postMessage` returns error of send
  * consumerclaim marks message only if POST was good, no error on post bubbled up

### DIFF
--- a/contrib/kafka/pkg/adapter/adapter.go
+++ b/contrib/kafka/pkg/adapter/adapter.go
@@ -79,8 +79,15 @@ func (a *Adapter) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.Co
 	for msg := range claim.Messages() {
 		logger.Info("Received: ", zap.Any("value", string(msg.Value)))
 
-		go a.postMessage(context.TODO(), msg)
-		sess.MarkMessage(msg, "")
+		go func(msg *sarama.ConsumerMessage) {
+			// send and mark message if post was successfull
+			if err := a.postMessage(context.TODO(), msg); err == nil {
+				sess.MarkMessage(msg, "")
+				logger.Debug("Successfully sent event to sink")
+			} else {
+				logger.Error("Sending event to sink failed: ", zap.Error(err))
+			}
+		}(msg)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


matzew:
@dubee mind taking a look ?

bbrowning:
Thank you sir!

PR #309, opened 2019-03-28
Add check_suite events added to githubsource types

iancoffey:
## Proposed Changes

Adds handling for check_suite events from github events API. This involved updating to GH API v3->v5.

```release-note

None
```


sebgoa:
Ideally we should add a test in https://github.com/knative/eventing-sources/blob/master/pkg/adapter/github/adapter_test.go

/ok-to-test

matzew:
+1 on adding a test for that

iancoffey:
Indeed, will add tests, ty!

iancoffey:
So this PR grew quite a bit - it turns the GH v3 api does not support `check_suite` events, so I upgraded this project to use Gh events v5.

Now, it supports check suites and has a test as well 😄

matzew:
Release notes ?   regarding the version upgrade ?

PR #311, opened 2019-03-28
Rename release artifacts to eventing-sources.yaml

matzew:
It was still `release.yaml`, but all others were like `build.yaml`, `serving.yaml` or `eventing.yaml`

+ update doc that were pointing to "old" `release.yaml` from eventing (not eventing-sources)

### DIFF
--- a/samples/awssqs_source/README.md
+++ b/samples/awssqs_source/README.md
@@ -10,7 +10,7 @@
 1.  If your installed version of Eventing did not include the in-memory channel
     provisioner, install the
     [in-memory `ClusterChannelProvisioner`](https://github.com/knative/eventing/tree/master/config/provisioners/in-memory-channel)
-    now. If you installed Eventing using the `release.yaml` file, the channel
+    now. If you installed Eventing using the `eventing.yaml` file, the channel
     provisioner was included. (See the
     [Custom install guide](https://www.knative.dev/docs/install/knative-custom-install/)
     for information about what is include in each install file.)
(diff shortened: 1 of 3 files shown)


evankanderson:
I think we eventually want to remove all the items from `release.yaml` in eventing-sources, and instead package each source individually (and probably move them all to `contrib/` as well.

I don't know if we have time for this for the 0.5.0 release, but we should pull it into a future release.

**Customer story**
_I want to be able to install specific event sources without running additional types on my cluster which I do not want (e.g. no AWS sources when I need cron)._

samples/awssqs_source/README.md
```diff
@@ -10,7 +10,7 @@
 1.  If your installed version of Eventing did not include the in-memory channel
     provisioner, install the
     [in-memory `ClusterChannelProvisioner`](https://github.com/knative/eventing/tree/master/config/provisioners/in-memory-channel)
-    now. If you installed Eventing using the `release.yaml` file, the channel
+    now. If you installed Eventing using the `eventing.yaml` file, the channel
```
evankanderson: Thanks for catching this -- looks like we were pointing at the wrong file since 0.4.0!

grantr: **No, this change is incorrect.** `eventing.yaml` does NOT include the in-memory channel provisioner. `release.yaml` does. https://github.com/knative/eventing/blob/18f7b078bdd9ea70a2b4db88abbedc321df0ed5c/hack/release.sh#L32

PR #315, opened 2019-03-29
Address function and variable naming lint errors

trshafer:
## Proposed Changes

  * Internal refactor

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controller/sinks/sinks_test.go
+++ b/pkg/controller/sinks/sinks_test.go
@@ -68,7 +68,7 @@ func TestGetSinkURI(t *testing.T) {
 		},
 		"nil hostname": {
 			objects: []runtime.Object{
-				getAddressable_nilHostname(),
+				getAddressableNilHostname(),
 			},
 			namespace: testNS,
 			ref:       getUnaddressableRef(),
@@ -76,7 +76,7 @@ func TestGetSinkURI(t *testing.T) {
 		},
 		"nil sink": {
 			objects: []runtime.Object{
-				getAddressable_nilHostname(),
+				getAddressableNilHostname(),
 			},
 			namespace: testNS,
 			ref:       nil,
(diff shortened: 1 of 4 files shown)


pkg/controller/sinks/sinks_test.go
```diff
@@ -151,23 +151,7 @@ func getAddressable_noStatus() *unstructured.Unstructured {
 	}
 }
 
-func getAddressable_nilAddress() *unstructured.Unstructured {
```
evankanderson: Was this not used?

evankanderson: If so, does that mean we are missing a test?

pkg/reconciler/containersource/containersource_test.go
```diff
@@ -637,7 +637,7 @@ func TestObjectHasDeleteTimestamp(t *testing.T) {
 	if diff := cmp.Diff(want, got); diff != "" {
 		t.Errorf("unexpected returned object (-want, +got) = %v", diff)
 	}
-	var wantErr error = nil
+	var wantErr error
 	if diff := cmp.Diff(wantErr, gotErr); diff != "" {
```
evankanderson: Why does this not just check that `gotErr == nil`?

pkg/controller/sinks/sinks_test.go
trshafer: The function was not called. Here's what that test would look like: https://github.com/knative/eventing-sources/pull/316. I can wait for that PR to be merged in and then fix the name change in this PR.

evankanderson:
(Otherwise, this removal will break #316 )

FWIW, I'd have just added the #316 change to this PR.

pkg/reconciler/containersource/containersource_test.go
trshafer: Done

PR #321, opened 2019-03-29
Add a ConcurrencyPerPartition option to KafkaSource

### LINKED ISSUE #320: Kafka Source creates unbounded goroutines on topics with many messages
## Expected Behavior
The Kafka Source should be able to consume messages from a Kafka topic with many thousands of messages without error.

## Actual Behavior
The Kafka Source creates a new goroutine for every message it consumes from every partition of a topic. If you dump a lot of messages into a topic, this ends up creating so many goroutines that you start running out of OS level resources - sockets, file descriptors, and such.

  1. Dump 50k messages into a Kafka topic.
  2. Create a KafkaSource pointing to that topic that sinks those messages to something that can handle the massive concurrent load about to be unleashed upon it.
  3. Watch the receive adapter pod start spewing messages after a bit as it stops being able to create new HTTP clients.

## Additional Info

bbrowning:
Fixes #320

## Proposed Changes

  * Add a spec.concurrencyPerPartition field to the KafkaSource CRD.

Copied from the 1st commit in this PR:

This field, if set, will limit that maximum number of concurrent sends (ie POSTS) of CloudEvent messages at a time for each partition. If left unset or set to 0, which is the default, the concurrency is unbounded.

When processing a large number of messages from a Kafka topic, the unbounded behavior leads to massive numbers of goroutines being created. This leads to problems with too many open sockets attempting to all POST to the sink at the same time.

So, in practice, we may want to actually default this to a reasonably high value that allows for a high throughput per topic partition without blowing up if processing a huge backlog of messages on a topic.

Note that, as the name implies, this bounds concurrency per partition. If a topic has many partitions and only a single receive adapter pod running (the default today) then you'll have up to `ConcurrencyPerPartition * numPartitions` concurrent HTTP POSTS to the sink.

### DIFF
--- a/contrib/kafka/cmd/receive_adapter/main.go
+++ b/contrib/kafka/cmd/receive_adapter/main.go
@@ -33,14 +33,15 @@ import (
 )
 
 const (
-	envBootstrapServers = "KAFKA_BOOTSTRAP_SERVERS"
-	envTopics           = "KAFKA_TOPICS"
-	envConsumerGroup    = "KAFKA_CONSUMER_GROUP"
-	envNetSASLEnable    = "KAFKA_NET_SASL_ENABLE"
-	envNetSASLUser      = "KAFKA_NET_SASL_USER"
-	envNetSASLPassword  = "KAFKA_NET_SASL_PASSWORD"
-	envNetTLSEnable     = "KAFKA_NET_TLS_ENABLE"
-	envSinkURI          = "SINK_URI"
+	envBootstrapServers        = "KAFKA_BOOTSTRAP_SERVERS"
+	envTopics                  = "KAFKA_TOPICS"
+	envConsumerGroup           = "KAFKA_CONSUMER_GROUP"
+	envNetSASLEnable           = "KAFKA_NET_SASL_ENABLE"
+	envNetSASLUser             = "KAFKA_NET_SASL_USER"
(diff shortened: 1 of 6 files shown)


bbrowning:
I did not add a release note for this. But, if KafkaSource makes it into 0.5 and this doesn't make it until 0.6 we'll want to add a release note.

We should also discuss defaulting this to some reasonable value instead of being unbounded. I left it unbounded to match the current behavior.

bbrowning:
cc @dubee @matzew

bbrowning:
/hold

After discussing this over Slack, @markfisher pointed out that there really shouldn't be a goroutine inside `ConsumeClaim` at all here. And that the knob to control consumer concurrency should be purely driven by the number of partitions in the Kafka topic, as is the usual Kafka consumer way. The goroutine inside `ConsumeClaim`, both with or without this PR, presents other problems as-written as far as committing of offsets and acking of events - the code assumes each offsets can be committed individually when in reality once you commit a specific offset it commits all messages with offsets less than or equal to that offset.

So, I'll take Mark's changes into account, removing the goroutine entirely, and see if that fixes the stability issues under load of the receive adapter.

matzew:
FWIW, I've filed a pr here: 
https://github.com/knative/eventing-sources/pull/322

feel free to close that, if you do update this :)

bbrowning:
Closing - this is the wrong solution

PR #322, opened 2019-03-30
Removing inner claim goroutine

matzew:
based on slack chats of @bbrowning and @markfisher I am removing the inner Claim `goroutine`. Matching the Sarama "best practises", by the documentation:
https://github.com/Shopify/sarama/blob/master/examples/consumergroup/main.go#L124

From the Sarama doc:
```
	// 3. For each of the assigned claims the handler's ConsumeClaim() function is then called
	//    in a separate goroutine which requires it to be thread-safe. Any state must be carefully protected
```
Perhaps we should add a comment in the code, to _not_ put a goroutine inside the `ConsumeClaim`?

Others might like to do the same failure, that I did: and overuse (or abuse) concurrency.

### DIFF
--- a/contrib/kafka/pkg/adapter/adapter.go
+++ b/contrib/kafka/pkg/adapter/adapter.go
@@ -79,15 +79,13 @@ func (a *Adapter) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.Co
 	for msg := range claim.Messages() {
 		logger.Info("Received: ", zap.Any("value", string(msg.Value)))
 
-		go func(msg *sarama.ConsumerMessage) {
-			// send and mark message if post was successful
-			if err := a.postMessage(context.TODO(), msg); err == nil {
-				sess.MarkMessage(msg, "")
-				logger.Debug("Successfully sent event to sink")
-			} else {
-				logger.Error("Sending event to sink failed: ", zap.Error(err))
-			}
-		}(msg)
+		// send and mark message if post was successful
+		if err := a.postMessage(context.TODO(), msg); err == nil {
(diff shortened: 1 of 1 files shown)


matzew:
> Perhaps we should add a comment in the code, to not put a goroutine inside the ConsumeClaim?

upstream like a note too: https://github.com/Shopify/sarama/pull/1334

so I will add such a note to our code ...

bbrowning:
Is this still a WIP? Based on all the community Kafka input via Slack Friday/weekend it sounds like this is the right approach.

matzew:
@bbrowning ha! I was about to reach out on this... removing the WIP

bbrowning:
I've manually tested this change on a real cluster under some actual load and can confirm this fixes #320 as well as aligns the Kafka Source concurrency story with Kafka itself - ie add more partitions to a topic to increase concurrency.

/lgtm

evankanderson:
Since this is a data loss bug, we should patch this back to the 0.5 and possibly 0.4 release branches.

grantr:
Was 0.5 already cut? If so then yes.

I haven't done a backport before, so I'm not sure how Prow will handle it, but here's how I imagine it working:
1. Create a branch
1. Cherry-pick 9ad60cc3a44abc75058ab2fbd2ce76318e65ac7a
1. Create a new PR with `release-0.4` as base

bbrowning:
The Kafka Source didn't exist in 0.4, did it? Or am I misreading git history?

matzew:
NOPE ! that thing is new to 0.5.0

did ping @evankanderson on slack last night about that already

PR #329, opened 2019-04-01
Stop making sensitive kafka information available for logging

### LINKED ISSUE #328: stop logging user information, raw messages in Kafka event-source
## Expected Behavior
Setting up the kafka event-source shouldn't leak user/password combinations as well as log actual message contents (imagine if that was sensitive content) outside of the intended sink.  General configuration and metadata makes sense to log.

## Actual Behavior
Running through the Apache Kafka [README](https://github.com/knative/eventing-sources/blob/master/contrib/kafka/samples/README.md) shows log files that display from the event-source:

`{"level":"info","ts":"2019-03-19T22:31:52.689Z","caller":"receive_adapter/main.go:97","msg":"Starting Apache Kafka Receive Adapter...","adapter":{"BootstrapServers":"...","Topic":"...","ConsumerGroup":"...","Net":{"SASL" {"Enable":true,"User":"...","Password":"..."},"TLS":{"Enable":true}},"SinkURI":"http://event-display.default.svc.cluster.local/"}}`

Likewise, later, the README shows logging of the actual messages themselves:
`{"level":"info","ts":1553034726.5351956,"logger":"fallback","caller":"adapter/adapter.go:121","msg":"Received: {value 15 0 {\"msg\": \"This is a test!\"} <nil>}"}`

This is due to logging mechanism in the receive_adapter using reflection to print out the entire adapter struct, and then again in the adapter itself, printing out the message.

  1. Follow steps specified in the Apache Kafka README
  2.
  3.

## Additional Info

lberk:
The adapter currently logs the kafka message directly to the console
instead; log metadata such as partition, topic, offset instead
The receive adapter is currently even worse due to the zap logging
done by reflection.  This prints out every struct parameter including
user/password combo.  Drop that and log less sensitive metadata.

Fixes #328

## Proposed Changes

  * Alter receive_adapter to print out a predefined set of field instead of reflecting the entire adapter struct
  * Change the adapter itself to log metadata only (partition, topic, offset)
  * Update the README to match the updated output

### DIFF
--- a/contrib/kafka/cmd/receive_adapter/main.go
+++ b/contrib/kafka/cmd/receive_adapter/main.go
@@ -95,10 +95,10 @@ func main() {
 	stopCh := signals.SetupSignalHandler()
 
 	logger.Info("Starting Apache Kafka Receive Adapter...",
-		zap.String("Bootstrap Server", string(adapter.BootstrapServers)),
-		zap.String("Topics", string(adapter.Topics)),
-		zap.String("ConsumerGroup", string(adapter.ConsumerGroup)),
-		zap.String("SinkURI", string(adapter.SinkURI)),
+		zap.String("bootstrapServer", adapter.BootstrapServers),
+		zap.String("Topics", adapter.Topics),
+		zap.String("ConsumerGroup", adapter.ConsumerGroup),
+		zap.String("SinkURI", adapter.SinkURI),
 		zap.Bool("TLS", adapter.Net.SASL.Enable))
 	if err := adapter.Start(ctx, stopCh); err != nil {
 		logger.Fatal("failed to start adapter: ", zap.Error(err))
--- a/contrib/kafka/cmd/receive_adapter/main.go
+++ b/contrib/kafka/cmd/receive_adapter/main.go
@@ -95,7 +95,7 @@ func main() {
 	stopCh := signals.SetupSignalHandler()
 
 	logger.Info("Starting Apache Kafka Receive Adapter...",
-		zap.String("bootstrapServer", adapter.BootstrapServers),
+		zap.String("bootstrap_server", adapter.BootstrapServers),
 		zap.String("Topics", adapter.Topics),
 		zap.String("ConsumerGroup", adapter.ConsumerGroup),
 		zap.String("SinkURI", adapter.SinkURI),
--- a/contrib/kafka/cmd/receive_adapter/main.go
+++ b/contrib/kafka/cmd/receive_adapter/main.go
@@ -94,7 +94,12 @@ func main() {
 
 	stopCh := signals.SetupSignalHandler()
 
-	logger.Info("Starting Apache Kafka Receive Adapter...", zap.Reflect("adapter", adapter))
+	logger.Info("Starting Apache Kafka Receive Adapter...",
+		zap.String("Bootstrap Server", string(adapter.BootstrapServers)),
+		zap.String("Topics", string(adapter.Topics)),
+		zap.String("ConsumerGroup", string(adapter.ConsumerGroup)),
+		zap.String("SinkURI", string(adapter.SinkURI)),
+		zap.Bool("TLS", adapter.Net.SASL.Enable))
 	if err := adapter.Start(ctx, stopCh); err != nil {
 		logger.Fatal("failed to start adapter: ", zap.Error(err))
 	}
--- a/contrib/kafka/pkg/adapter/adapter.go
+++ b/contrib/kafka/pkg/adapter/adapter.go
@@ -77,8 +77,9 @@ func (a *Adapter) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.Co
 	logger := logging.FromContext(context.TODO())
 
 	for msg := range claim.Messages() {
-		logger.Info("Received: ", zap.Any("value", string(msg.Value)))
-
+		logger.Info("Received: ", zap.String("topic", string(msg.Topic)),
+			zap.Int("partition", int(msg.Partition)),
+			zap.Int("offset:", int(msg.Offset)))
 		go func(msg *sarama.ConsumerMessage) {
 			// send and mark message if post was successful
 			if err := a.postMessage(context.TODO(), msg); err == nil {
--- a/contrib/kafka/pkg/adapter/adapter.go
+++ b/contrib/kafka/pkg/adapter/adapter.go
@@ -77,8 +77,8 @@ func (a *Adapter) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.Co
 	logger := logging.FromContext(context.TODO())
 
 	for msg := range claim.Messages() {
-		logger.Info("Received: ", zap.String("topic", string(msg.Topic)),
-			zap.Int("partition", int(msg.Partition)),
+		logger.Info("Received: ", zap.String("topic:", string(msg.Topic)),
+			zap.Int("partition:", int(msg.Partition)),
 			zap.Int("offset:", int(msg.Offset)))
 		go func(msg *sarama.ConsumerMessage) {
 			// send and mark message if post was successful
--- a/contrib/kafka/pkg/adapter/adapter.go
+++ b/contrib/kafka/pkg/adapter/adapter.go
@@ -77,9 +77,9 @@ func (a *Adapter) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.Co
 	logger := logging.FromContext(context.TODO())
 
 	for msg := range claim.Messages() {
-		logger.Info("Received: ", zap.String("topic:", string(msg.Topic)),
-			zap.Int("partition:", int(msg.Partition)),
-			zap.Int("offset:", int(msg.Offset)))
+		logger.Debug("Received: ", zap.String("topic:", msg.Topic),
+			zap.Int32("partition:", msg.Partition),
+			zap.Int64("offset:", msg.Offset))
 		go func(msg *sarama.ConsumerMessage) {
 			// send and mark message if post was successful
 			if err := a.postMessage(context.TODO(), msg); err == nil {
--- a/contrib/kafka/samples/README.md
+++ b/contrib/kafka/samples/README.md
@@ -117,7 +117,7 @@ and an Event Display Service.
    ```
    $ kubectl logs kafka-source-xlnhq-5544766765-dnl5s
    ...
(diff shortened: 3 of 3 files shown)


contrib/kafka/cmd/receive_adapter/main.go
```diff
@@ -94,7 +94,12 @@ func main() {
 
 	stopCh := signals.SetupSignalHandler()
 
-	logger.Info("Starting Apache Kafka Receive Adapter...", zap.Reflect("adapter", adapter))
+	logger.Info("Starting Apache Kafka Receive Adapter...",
+		zap.String("Bootstrap Server", string(adapter.BootstrapServers)),
+		zap.String("Topics", string(adapter.Topics)),
+		zap.String("ConsumerGroup", string(adapter.ConsumerGroup)),
+		zap.String("SinkURI", string(adapter.SinkURI)),
+		zap.Bool("TLS", adapter.Net.SASL.Enable))
```
matzew: :+1: nice find!

contrib/kafka/pkg/adapter/adapter.go
```diff
@@ -77,8 +77,9 @@ func (a *Adapter) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.Co
 	logger := logging.FromContext(context.TODO())
 
 	for msg := range claim.Messages() {
-		logger.Info("Received: ", zap.Any("value", string(msg.Value)))
-
+		logger.Info("Received: ", zap.String("topic", string(msg.Topic)),
+			zap.Int("partition", int(msg.Partition)),
+			zap.Int("offset:", int(msg.Offset)))
```
matzew: minor:
`offset:` ends with a `:`  - should have the same for `topic` and `partition` ?

contrib/kafka/samples/README.md
```diff
@@ -103,7 +103,7 @@ and an Event Display Service.
    configuration.
    ```
    $ kubectl logs kafka-source-xlnhq-5544766765-dnl5s
-   {"level":"info","ts":"2019-03-19T22:31:52.689Z","caller":"receive_adapter/main.go:97","msg":"Starting Apache Kafka Receive Adapter...","adapter":{"BootstrapServers":"...","Topic":"...","ConsumerGroup":"...","Net":{"SASL":{"Enable":true,"User":"...","Password":"..."},"TLS":{"Enable":true}},"SinkURI":"http://event-display.default.svc.cluster.local/"}}
+   {"level":"info","ts":"2019-04-01T19:09:32.164Z","caller":"receive_adapter/main.go:97","msg":"Starting Apache Kafka Receive Adapter...","Bootstrap Server":"...","Topics":".","ConsumerGroup":"...","SinkURI":"http://event-display.default.svc.cluster.local/","TLS":false}
```
matzew: :+1:

matzew:
@evankanderson @grantr given that the current IMPL logs sensible data, I'd recommend getting this into 0.5  branch/tag

contrib/kafka/pkg/adapter/adapter.go
lberk: Ah, nice catch.  Thanks for the review, I've pushed a commit that fixes this and updates the README accordingly!

grantr: These should be `topic`, `partition`, and `offset`. No need for the `:` separator - zap's formatter will take care of that.

contrib/kafka/cmd/receive_adapter/main.go
```diff
@@ -94,7 +94,12 @@ func main() {
 
 	stopCh := signals.SetupSignalHandler()
 
-	logger.Info("Starting Apache Kafka Receive Adapter...", zap.Reflect("adapter", adapter))
+	logger.Info("Starting Apache Kafka Receive Adapter...",
+		zap.String("Bootstrap Server", string(adapter.BootstrapServers)),
```
grantr: Remove the space here. That makes the field names easier to use in search queries.

@yanweiguo do you have any guidance on casing for field names in logs? Should this be `BootstrapServer`, `bootstrapServer`, or `bootstrap_server`? I think I prefer `bootstrapServer` or `bootstrap_server`.

contrib/kafka/pkg/adapter/adapter.go
```diff
@@ -77,8 +77,9 @@ func (a *Adapter) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.Co
 	logger := logging.FromContext(context.TODO())
 
 	for msg := range claim.Messages() {
-		logger.Info("Received: ", zap.Any("value", string(msg.Value)))
-
+		logger.Info("Received: ", zap.String("topic:", string(msg.Topic)),
```
grantr: No need for the casts here. `msg.Topic`, `msg.Partition`, and `msg.Offset` are already the expected types.

grantr: I think this should be `Debug` logging. Logging every message at `Info` can cause performance issues at high throughput.

contrib/kafka/samples/README.md
```diff
@@ -117,7 +117,7 @@ and an Event Display Service.
    ```
    $ kubectl logs kafka-source-xlnhq-5544766765-dnl5s
    ...
-   {"level":"info","ts":1553034726.5351956,"logger":"fallback","caller":"adapter/adapter.go:121","msg":"Received: {value 15 0 {\"msg\": \"This is a test!\"} <nil>}"}
+   {"level":"info","ts":1554145778.9344022,"logger":"fallback","caller":"adapter/adapter.go:80","msg":"Received: {topic: 15 0 ... <nil>} {partition: 11 2  <nil>} {offset: 11 0  <nil>}"}
```
grantr: Is this what the log messages look like? I wouldn't expect go structs to be logged here, given the log call at https://github.com/knative/eventing-sources/pull/329/files#diff-ca72cd4e88c155c7eabbb1bbd3d55b2fR80.

I'd expect something like this:

```
{"level":"info","ts":1554145778.9344022,"logger":"fallback","caller":"adapter/adapter.go:80",
"msg":"Received","topic":"mytopic","partition":11,"offset":151}
```

contrib/kafka/cmd/receive_adapter/main.go
lberk: I'll use your suggested bootstrapServer for now (also keeping relatively inline with the conventions inside the receive_adapter) unless we have a more preferred logging convention specified elsewhere.

contrib/kafka/pkg/adapter/adapter.go
lberk: Sure, addressed in upcoming commit

lberk: Ok, Partition variable are int32's and Offset is int64.  Would you consider it acceptable to use zap.Int32() and zap.Int64() for the respective variables instead of the cast?

grantr: Yep that's what I'd prefer!

contrib/kafka/cmd/receive_adapter/main.go
lberk: I'll also remove the casts here as well

matzew: FWIW, the CCP, over in eventing repo, uses `bootstrap_servers` in the config map for config:

https://github.com/knative/eventing/blob/master/contrib/kafka/config/kafka.yaml#L111

yanweiguo: [Here](https://github.com/knative/pkg/blob/master/logging/logkey/constants.go) are the log keys we are currently using in logs. I don't think we have a strict naming policy for logs currently. Personally I prefer `bootstrap_server`.

cc @mdemirhan

contrib/kafka/samples/README.md
lberk: Looked into this a bit now.  Yes, that's what the logs look like.  It appears that both prior logging messages and current ones are using the fallback logger (which is why we have the extra output).

Testing this quickly locally by commenting out the 
`logger := logging.FromContext(context.TODO())`
line and adding in
```
logger := zap.NewExample()
defer logger.Sync()
```
Seems to produce the logging format you describe.

May I suggest that, due to this being a preexisting issue, we leave it for another, more focused Issue/PR regarding the use of the fallback logger?

contrib/kafka/cmd/receive_adapter/main.go
lberk: Sure, changed to `bootstrap_server` in e5053ba

contrib/kafka/cmd/receive_adapter/main.go
```diff
@@ -95,7 +95,7 @@ func main() {
 	stopCh := signals.SetupSignalHandler()
 
 	logger.Info("Starting Apache Kafka Receive Adapter...",
-		zap.String("bootstrapServer", adapter.BootstrapServers),
+		zap.String("bootstrap_server", adapter.BootstrapServers),
```
aslom: Underlying yaml parameter name is `bootstrapServers` (see it is plural with "s" at the end) I think it would be good to keep it consistent for people that will search in logs when troubleshooting?
https://github.com/knative/eventing-sources/blob/master/contrib/kafka/config/300-kafkasource.yaml#L46

contrib/kafka/cmd/receive_adapter/main.go
```diff
@@ -95,7 +95,7 @@ func main() {
...
 		zap.String("Topics", adapter.Topics),
 		zap.String("ConsumerGroup", adapter.ConsumerGroup),
```
aslom: Similarly I would keep the names of parameters exactly the same as far as capitalization goes: `topics`, `consumerGroup` etc.

```
		zap.String("consumerGroup", adapter.ConsumerGroup),
```

contrib/kafka/cmd/receive_adapter/main.go
lberk: This was already discussed here: https://github.com/knative/eventing-sources/pull/329#discussion_r271031783

contrib/kafka/cmd/receive_adapter/main.go
lberk: This was already discussed: https://github.com/knative/eventing-sources/pull/329#discussion_r271031783

contrib/kafka/samples/README.md
grantr: Sure, I created #332.

PR #334, opened 2019-04-02
Remove message dumper.

n3wscott:
Breaks the release.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -24,7 +24,6 @@ COMPONENTS=(
   ["event-display.yaml"]="config/tools/event-display"
   ["camel.yaml"]="contrib/camel/config"
   ["kafka.yaml"]="contrib/kafka/config"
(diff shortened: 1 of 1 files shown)


vaikas-google:
This is only from the release, perhaps update the subject of the PR to reflect that. If our eventual plan is to remove in favor of something else for example, we probably still have a ton of documentation that needs to be updated.

grantr:
I believe that already occurred in https://github.com/knative/eventing-sources/pull/249. This is just fixing the release to not try building a missing artifact.

n3wscott:
> This is only from the release, perhaps update the subject of the PR to reflect that. If our eventual plan is to remove in favor of something else for example, we probably still have a ton of documentation that needs to be updated.

It turns out this was fixing a bad merge: https://github.com/knative/eventing-sources/pull/264/files#diff-cede583a27c227d458ca67079957c8a2R27

PR #337, opened 2019-04-03
Adding a release.yaml

matzew:
In eventing-sources we had a `release.yaml` that now the `eventing-sources`, to match other projects (see #311)

It might be handy to actually have a `release.yaml` that combines `eventing-sources` and the handy `event-display` 

/cc @n3wscott 

Perhaps for 0.5.1 ?

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -27,12 +27,30 @@ COMPONENTS=(
 )
 readonly COMPONENTS
 
+declare -A RELEASES
+RELEASES=(
+  ["release.yaml"]="eventing-sources.yaml event-display.yaml"
+)
+readonly RELEASES
+
 function build_release() {
+  # Build the components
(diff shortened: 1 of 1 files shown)


n3wscott:
/hold

I do not foresee wide adoption of a single installable of eventing sources, they are suppose to be opt in single components.

There is a larger plan to pull the fundamental building blocks out of this repo and move them back to eventing. (like container source and anything else that is fundamental). Given this, I don't think there is a case for installing all sources in a single yaml.

matzew:
> There is a larger plan to pull the fundamental building blocks out of this repo and move them back to eventing. (like container source and anything else that is fundamental). 

Oh, that's actually a nice idea ... installing ContainerSource and the manager to the "eventing core" (that's how I call it)

PR #345, opened 2019-04-05
Update dep for test-infra

nbarthwal:
## Proposed Changes

  * Updating test-infra dependency of vendor using dep update


adrcunha:
This is doing more than it should: 190 changed files.

/hold

adrcunha:
> /retest

This won't work, this update is wrong. Your Gopkg file has 100+ changes, when it's expected that only 2 lines change (only test-infra).

nbarthwal:
This is faulty, will send a new PR!

PR #350, opened 2019-04-06
Format markdown

mattmoor-sockpuppet:
Produced via: `prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github)`


lberk:
Considering the frequency of these kind of commits (and noticing I'm the cause of least one of them myself).  Would it be worth adding your script to a pre-commit hook and noting that for potential contributors in the various knative repos?

PR #289, opened 2019-03-26
Change brokers references to bootstrap servers for Kafka event source

dubee:
Replaces references to `brokers` with `bootstrap servers` for consistency with Apache Kafka documentation.

PR #351, opened 2019-04-08
contrib/kafka: use secrets for SASL user and password

danp:
## Proposed Changes

Change the Kafka source spec to use SecretValueFromSource for user and
password, similar to the GitHub source.

This means the SASL user and password will no longer be stored in cleartext in the KafkaSource resources.

Based on #289 I did not allow for a graceful transition to this new scheme but happy to add it if needed.

**Release Note**

```release-note
KafkaSource User and Password are now references to keys in secrets instead of plaintext values. Existing KafkaSource resources will need to be updated.
```

### DIFF
--- a/contrib/kafka/config/300-kafkasource.yaml
+++ b/contrib/kafka/config/300-kafkasource.yaml
@@ -52,10 +52,37 @@ spec:
             consumerGroup:
               type: string
               minLength: 1
+            consumerGroup:
+              type: string
+            net:
+              properties:
+                sasl:
+                  properties:
+                    enable:
+                      type: boolean
+                    password:
+                      properties:
+                        secretKeyRef:
+                          type: object
+                      type: object
+                    user:
+                      properties:
+                        secretKeyRef:
+                          type: object
+                      type: object
+                  type: object
+                tls:
+                  properties:
+                    enable:
+                      type: boolean
+                  type: object
+              type: object
             serviceAccountName:
               type: string
             sink:
               type: object
+            topics:
+              type: string
           type: object
           required:
               - bootstrapServers
@@ -83,6 +110,9 @@ spec:
                 - status
                 type: object
               type: array
+            observedGeneration:
+              format: int64
+              type: integer
             sinkUri:
               type: string
           type: object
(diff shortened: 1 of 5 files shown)


danp:
Wasn't sure how to nicely regenerate `config/300-kafkasource.yaml` so I did roughly:

```
cd contrib/kafka
go run ../../vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go all
mv config/crds/sources_v1alpha1_kafkasource.yaml config/300-kafkasource.yaml
# cut out non-relevant changes, add back license header, etc
```

If there's a better way, please let me know!

matzew:
Mind adding a section on the readme for this ?

danp:
Which one were you thinking?

Did realize I need to update `contrib/kafka/samples/event-source.yaml`, will do that shortly.

matzew:
yeah, thanks @danp the readme in the samples and the related yamls doc

matzew:
IMO it would be nice if some of the new secret bits are captured here:
* https://github.com/knative/eventing-sources/tree/810ad72018e4878a20508eb8350a1e8584182f62/contrib/kafka/samples

Also, the `docs` repo has some high level details on the KafkaSource:
* https://github.com/knative/docs/pull/1099/files

I wonder what's best ... if we move the samples/docs to the `docs` repo and write some real doc for it, explaining all possible config settings ?

Thoughts ?

@samodell @evankanderson @abrennan89

danp:
Ok! I did update the sample YAML in https://github.com/knative/eventing-sources/pull/351/commits/810ad72018e4878a20508eb8350a1e8584182f62.

Should we consider further doc changes as part of future PRs? For example, the current sample README doesn't mention SASL at all.

matzew:
@danp I personally would like to see some more doc here.

I think in the long run we want to have some decent doc for all the sources, including all possible options.

In the long run, I'd like to see all docs mostly in the docs repo, and some short/dev references from _here_ to the docs repo.   

Perhaps @samodell  from the docs team has some thoughts here too /cc @abrennan89

matzew:
@danp btw. mind to rebase, when adding the doc ? 

there is a conflict here

matzew:
IMO it's good enough to release note that the `User` and `Password` are now references to secrets, instead of not so secure plaintext :joy: 

/cc @grantr

PR #353, opened 2019-04-08
Update Eventing Sources to use pkg with added context.

n3wscott:
## Proposed Changes

  * Update our usage of pkg to use the next context added in the webhook calls.
  * Dep on serving 0.5.0 

**Release Note**

```release-note
NONE.
```


Gopkg.toml
```diff
@@ -39,8 +39,8 @@ required = [
 # This controls when we upgrade apis independently of Serving.
 [[override]]
   name = "github.com/knative/pkg"
```
vaikas-google: Is there a reason we don't just bring this to current HEAD?

n3wscott: it is a moving target

n3wscott: I am matching what eventing is using.

n3wscott: you convinced me, updating...

PR #357, opened 2019-04-15
Missing RBAC rules for finalizers

matzew:
At least on OpenShift, which has fairly restrictive RBAC out of the
box, the sources are missing a RBAC rules that
requires permissions on the different *sources/finalizers resource.

See also #318

/cc @evankanderson

### DIFF
--- a/config/201-clusterrole.yaml
+++ b/config/201-clusterrole.yaml
@@ -39,9 +39,19 @@ rules:
 - apiGroups:
   - sources.eventing.knative.dev
   resources:
-  - '*/finalizers'
-  verbs:
-  - '*'
+  - awssqssources/finalizers
+  - containersources/finalizers
+  - cronjobsources/finalizers
+  - githubsources/finalizers
+  - kuberneteseventsources/finalizers
+  verbs: &everything
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
 
 # Source statuses update
 - apiGroups:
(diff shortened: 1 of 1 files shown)


config/201-clusterrole.yaml
```diff
@@ -35,6 +35,14 @@ rules:
   - patch
   - delete
 
+# Sources finalizer
+- apiGroups:
+  - sources.eventing.knative.dev
+  resources:
+  - '*/finalizers'
+  verbs:
```
n3wscott: `verbs: *everything`

config/201-clusterrole.yaml
```diff
@@ -35,6 +35,14 @@ rules:
   - patch
   - delete
 
+# Sources finalizer
+- apiGroups:
+  - sources.eventing.knative.dev
+  resources:
+  - '*/finalizers'
```
n3wscott: there is not another example of * in all of knative, I think we should be explicit.

matzew:
@n3wscott I am explicit now :-)

config/201-clusterrole.yaml
```diff
@@ -35,6 +35,24 @@ rules:
...
+# Sources finalizer
+- apiGroups:
+  - sources.eventing.knative.dev
+  resources:
+  - awssqssources/finalizers
+  - containersources/finalizers
+  - cronjobsources/finalizers
+  - githubsources/finalizers
+  - kuberneteseventsources/finalizers
+  verbs: &everything
```
n3wscott: change this to `verbs: *everything`

you are redefining the magic `everything` yaml tag

PR #360, opened 2019-04-16
Update event-sources to not use fallback logger

### LINKED ISSUE #332: Kafka adapter uses the fallback logger
Looked into this a bit now.  Yes, that's what the logs look like.  It appears that both prior logging messages and current ones are using the fallback logger (which is why we have the extra output).

Testing this quickly locally by commenting out the 
`logger := logging.FromContext(context.TODO())`
line and adding in
```
logger := zap.NewExample()
defer logger.Sync()
```
Seems to produce the logging format you describe.

May I suggest that, due to this being a preexisting issue, we leave it for another, more focused Issue/PR regarding the use of the fallback logger?

_Originally posted by @lberk in https://github.com/knative/eventing-sources/pull/329/files#r271099097_

lberk:
Fixes #332

Currently, event-sources don't properly set, or receive their loggers
from context.  Change this to set the initial logger in both the
event-source, as well as the underlying receiver.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,7 +18,6 @@ package main
 
 import (
 	"log"
-	"os"
 
 	"github.com/knative/eventing-sources/pkg/apis"
 	_ "github.com/knative/eventing-sources/pkg/apis/serving/v1alpha1"
@@ -32,15 +31,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 )
 
-const (
-	ConfigMapNameEnv = "CONFIG_LOGGING_NAME"
-)
-
 func main() {
 	logCfg := zap.NewProductionConfig()
 	logCfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
 	logger, err := logCfg.Build()
-	logger = logger.With(zap.String(logkey.ControllerType, os.Getenv(ConfigMapNameEnv)))
+	logger = logger.With(zap.String(logkey.ControllerType, "controller-manager"))
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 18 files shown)


pkg/controller/sdk/reconciler.go
```diff
@@ -46,8 +47,8 @@ var _ reconcile.Reconciler = &Reconciler{}
 // Reconcile compares the actual state with the desired, and attempts to
 // converge the two.
 func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	ctx := context.TODO()
-	logger := logging.FromContext(ctx)
+	ctx := logging.WithLogger(context.TODO(), r.logger.With(zap.Any("request", request)))
+	logger := r.logger
```
Harwayne: Shouldn't this be `logger := logging.FromContext(ctx)`? As-is, you are missing the `request` parameter added on the previous line.

cmd/manager/main.go
```diff
@@ -18,17 +18,33 @@ package main
...
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 )
 
+const (
+	ConfigMapNameEnv = "CONFIG_LOGGING_NAME"
```
Harwayne: Is this env var set anywhere?

pkg/controller/sdk/reconciler.go
lberk: Fixed in commit b18e66e

cmd/manager/main.go
lberk: Fixed in commit b18e66e and dropped the various CONFIG_LOGGING_NAME references, just set the string directly based on the source.

PR #362, opened 2019-04-17
use cloudevent alias pkg

matzew:
as per title

/cc @n3wscott

### DIFF
--- a/pkg/kncloudevents/good_client.go
+++ b/pkg/kncloudevents/good_client.go
@@ -1,26 +1,29 @@
 package kncloudevents
 
 import (
+	"github.com/cloudevents/sdk-go"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/client"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport/http"
 )
 
 func NewDefaultClient(target ...string) (client.Client, error) {
-	tOpts := []http.Option{http.WithBinaryEncoding()}
+	tOpts := []http.Option{cloudevents.WithBinaryEncoding()}
 	if len(target) > 0 && target[0] != "" {
-		tOpts = append(tOpts, http.WithTarget(target[0]))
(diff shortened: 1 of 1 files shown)


pkg/kncloudevents/good_client.go
```diff
@@ -1,26 +1,29 @@
 package kncloudevents
 
 import (
+	"github.com/cloudevents/sdk-go"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/client"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport/http"
 )
 
 func NewDefaultClient(target ...string) (client.Client, error) {
```
n3wscott: This client too in the sig

pkg/kncloudevents/good_client.go
```diff
@@ -1,26 +1,29 @@
 package kncloudevents
 
 import (
+	"github.com/cloudevents/sdk-go"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/client"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport/http"
```
n3wscott: Delete these imports

pkg/kncloudevents/good_client.go
```diff
@@ -1,26 +1,29 @@
...
-	t, err := http.New(tOpts...)
+	t, err := cloudevents.NewHTTPTransport(tOpts...)
 	if err != nil {
 		return nil, err
 	}
+
 	// Use the transport to make a new CloudEvents client.
-	c, err := client.New(t,
+	c, err := cloudevents.NewClient(t,
 		client.WithUUIDs(),
```
n3wscott: This can come from cloudevents too

pkg/kncloudevents/good_client.go
matzew: @n3wscott  `Option` is still in `http`,

see:  https://github.com/cloudevents/sdk-go/pull/116

PR #365, opened 2019-04-18
Fixing address in GitHub Receive Adapter

nachocano:
The github receive adapter is not working. 
It has been broken for ~2 weeks now, since https://github.com/knative/eventing-sources/pull/309
Fortunately, that change didn't make it to the release.

**Release Note**

```release-note
NONE
```


nachocano:
I came up with this while trying to add the CloudEvents subject to the GH source.
Unfortunately, this is the second time I find that the adapter is broken in a short period of time, which makes me wonder if we should actually move this away from here, or at least put more effort on e2e tests.

Harwayne:
> Unfortunately, this is the second time I find that the adapter is broken in a short period of time, which makes me wonder if we should actually move this away from here, or at least put more effort on e2e tests.

I agree we need tests for sources. Ideally e2e tests. But if we can't do that, then at least a simple shell script that can be run manually before submission (with the submitter stating they ran before submission).

For GitHub in particular, it seems like we should be able to create a test repo and store an API key to allow us to automate e2e tests against.

nachocano:
> with the submitter stating they ran before submission

Yes, I agree! At least thee submitter should state she/he ran the code

PR #379, opened 2019-04-29
Use gke istio add-on.

n3wscott:
Do not depend on serving istio.


adrcunha:
This PR shouldn't be necessary anymore, plus it would break local testing with non-GKE clusters. Updating test-infra to the latest version should be enough; @chaodaig and @Fredy-Z to confirm.

/hold

chaodaiG:
> This PR shouldn't be necessary anymore, plus it would break local testing with non-GKE clusters. Updating test-infra to the latest version should be enough; @chaodaiG and @Fredy-Z to confirm.
> 
> /hold

agree

PR #395, opened 2019-05-06
remove duplicated keys

matzew:
Follow up from @danp' #351  

the generation did re-order keys, and they got accidentally added (as dup)


danp:
👍, apologies

PR #397, opened 2019-05-06
update test-infra to remove dependency on Istio published by serving

chaodaiG:
Install Istio from Addon instead of consuming from serving publish


chaodaiG:
Yes this should not affect running e2e test locally, unless someone use e2e-tests.sh setting up a fresh cluster. 
And this also affects this: https://testgrid.knative.dev/knative-eventing-sources-0.5#continuous

PR #398, opened 2019-05-06
Adding transformer field to GCPPubSubSource CRD.

nachocano:
## Proposed Changes

  * Adding an optional addressable to GCPPubSubSource CRD called transformer to be able to transform GCP events.

## Pending
  * Default transformer. 
  * Cleanup "ce-type" logic.

**Release Note**

```release-note
GCPPubSubSource includes an optional `transformer` Addressable field that can be configured to do custom CloudEvents transformations.
```


nachocano:
/cc @Harwayne @vaikas-google 
still WIP but it would be nice if you can start looking at it...

contrib/gcppubsub/cmd/receive_adapter/main.go
```diff
@@ -69,6 +72,7 @@ func main() {
 		TopicID:        getRequiredEnv(envTopic),
 		SinkURI:        getRequiredEnv(envSinkURI),
 		SubscriptionID: getRequiredEnv(envSubscription),
+		TransformerURI: getRequiredEnv(envTransformerURI),
```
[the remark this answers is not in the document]
nachocano: required but can be empty... we are always setting it but might be just ""
I'm making some changes, and removing my dummy transformer. Give me few mins...

nachocano:
Removed  the default transformer for now until we further agree how it should behave.
Just added the transformer addressable field to GCPPubSubSource.
Ready to review this part.

contrib/gcppubsub/cmd/receive_adapter/main.go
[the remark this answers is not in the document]
nachocano: Does that make sense?

nachocano:
Added back a default transformer that does nothing for now...

contrib/gcppubsub/pkg/adapter/adapter.go
```diff
@@ -98,23 +111,36 @@ func (a *Adapter) receiveMessage(ctx context.Context, m PubSubMessage) {
...
+	event.SetID(m.ID())
+	event.SetTime(m.PublishTime())
+	event.SetDataContentType(*cloudevents.StringOfApplicationJSON())
+	event.SetSource(a.source)
+	event.SetData(m.Message())
+
 	// TODO: this will break when the upstream sender updates cloudevents versions.
 	// The correct thing to do would be to convert the message to a cloudevent if it is one.
 	et := sourcesv1alpha1.GcpPubSubSourceEventType
 	if override, ok := m.Message().Attributes[eventTypeOverrideKey]; ok {
```
nachocano: IMO we should remove this asap. If a particular source wants to do something like this (e.g., the gcs source), it should actually do it in the transformer.

contrib/gcppubsub/cmd/receive_adapter/main.go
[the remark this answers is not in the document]
nachocano: Moved to kelseyhightower. Thanks!

contrib/gcppubsub/config/300-gcppubsubsource.yaml
```diff
@@ -51,6 +51,8 @@ spec:
               type: string
             sink:
               type: object
+            transformer:
```
vaikas-google: Could you please add description fields for all these while you're here? Something like this:
https://github.com/vaikas-google/gcs/blob/master/config/gcs.yaml#L36

contrib/gcppubsub/pkg/adapter/adapter.go
```diff
@@ -98,23 +111,36 @@ func (a *Adapter) receiveMessage(ctx context.Context, m PubSubMessage) {
...
+		resp, err := a.transformerClient.Send(ctx, event)
+		if err != nil {
+			logger.Errorf("error transforming cloud event %q", event.ID())
+			return err
+		}
+		if resp != nil {
+			// Update the event with the transformed one.
+			event = *resp
+		} else {
+			logger.Warnf("cloud event %q was not transformed", event.ID())
```
vaikas-google: Should this return an error?

contrib/gcppubsub/cmd/receive_adapter/main.go
```diff
@@ -64,11 +58,17 @@ func main() {
...
 	adapter := &gcppubsub.Adapter{
-		ProjectID:      getRequiredEnv(envProject),
-		TopicID:        getRequiredEnv(envTopic),
-		SinkURI:        getRequiredEnv(envSinkURI),
-		SubscriptionID: getRequiredEnv(envSubscription),
+		ProjectID:      env.Project,
+		TopicID:        env.Topic,
+		SinkURI:        env.Sink,
+		SubscriptionID: env.Subscription,
+		TransformerURI: env.Transformer,
```
vaikas-google: Should we at least validate the URI to make sure it's a valid URI?

nachocano: we are not doing that for the sinkURI. I guess if it's not valid it will break when trying to create the httpClient. And it can be empty right? in that case we do not create any client...

contrib/gcppubsub/config/300-gcppubsubsource.yaml
nachocano: Done, thanks!

contrib/gcppubsub/pkg/adapter/adapter.go
nachocano: I was leaving it as in case the transformer does not populate properly the response. If so, then we at least send the //google.pubsub.topic.publish event. As we might not control the transformer code, I'm assuming it can be wrong.
Do you prefer to return an error instead?

contrib/gcppubsub/cmd/receive_adapter/main.go
```diff
@@ -64,11 +58,17 @@ func main() {
 		log.Fatalf("Unable to create logger: %v", err)
 	}
 
+	var env envConfig
+	if err := envconfig.Process("", &env); err != nil {
+		log.Fatal("Failed to process env var", zap.Error(err))
```
Harwayne: Is this a sugared logger? If so, it should be `Fatalw` (my preference is just switch to a desugared logger to make sure these bugs don't occur).

contrib/gcppubsub/config/300-gcppubsubsource.yaml
```diff
@@ -45,14 +45,22 @@ spec:
           properties:
             gcpCredsSecret:
               type: object
+              description: "Credential to use to poll the GCP PubSub Subscription. It is not used to create or delete the Subscription, only to poll it. The value of the secret entry must be a service account key in the JSON format (see https://cloud.google.com/iam/docs/creating-managing-service-account-keys)."
             googleCloudProject:
               type: string
+              description: "ID of the Google Cloud Project that the PubSub Topic exists in."
```
Harwayne: I would add an example here, as historically GCP project name vs ID vs display name has been very confusing.

description: "ID of the Google Cloud Project that the PubSub Topic exists in. E.g. 'my-project-1234' rather than its display name, 'My Project' or its number '1234567890'."

contrib/gcppubsub/config/300-gcppubsubsource.yaml
```diff
@@ -45,14 +45,22 @@ spec:
...
             serviceAccountName:
               type: string
+              description: "Name of the ServiceAccount that will be used to run the Receive Adapter Deployment."
```
Harwayne: Give an example. Is this the full email address? Just the name before the '@'?

contrib/gcppubsub/config/300-gcppubsubsource.yaml
Harwayne: This is awesome, thanks for doing it!

contrib/gcppubsub/pkg/adapter/adapter.go
Harwayne: I agree.

contrib/gcppubsub/pkg/adapter/adapter.go
Harwayne: I like the idea that the transformer can hide events. For example, events that are not on an allowed list or those that do not pass schema validation. I think our 'canonical' filtering of events is to return a 2XX with an empty body.

So I wouldn't return an error here, I would just `return nil`.

contrib/gcppubsub/pkg/apis/sources/v1alpha1/gcp_pubsub_types.go
```diff
@@ -83,7 +87,7 @@ const (
 )
 
 // GetGcpPubSubSource returns the GcpPubSub CloudEvent source value.
```
Harwayne: GetGcpPubSubSource -> GcpPubSubSource

contrib/gcppubsub/pkg/apis/sources/v1alpha1/gcp_pubsub_types.go
```diff
@@ -146,11 +157,26 @@ func (s *GcpPubSubSourceStatus) MarkSink(uri string) {
 	}
 }
 
+// MarkSink sets the condition that the source has a transformer configured.
+func (s *GcpPubSubSourceStatus) MarkTransformer(uri string) {
+	s.TransformerURI = uri
+	if len(uri) > 0 {
+		gcpPubSubSourceCondSet.Manage(s).MarkTrue(GcpPubSubConditionTransformerProvided)
+	} else {
+		gcpPubSubSourceCondSet.Manage(s).MarkUnknown(GcpPubSubConditionTransformerProvided, "TransformerEmpty", "Transformer has resolved to empty.%s", "")
```
Harwayne: Why the trailing "%s" that has an empty string arg?

PR #401, opened 2019-05-07
golang format tools

mattmoor-sockpuppet:
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -type f -name '*.go' -print))`
  `goimports -w $(find -name '*.go' | grep -v vendor)`

### DIFF
--- a/pkg/adapter/github/adapter.go
+++ b/pkg/adapter/github/adapter.go
@@ -24,7 +24,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/cloudevents/sdk-go"
+	cloudevents "github.com/cloudevents/sdk-go"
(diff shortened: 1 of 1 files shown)


rootfs:
/lgtm

It would be nice to have the CI to check them

evankanderson:
-1 to running the `gofmt` tools being a presubmit blocker -- I'd rather do post-cleanup here than force everyone to do an extra submit (and possibly re-lgtm, etc) for something which can be doen by an automated tool.

At some point in the future, sockpuppet may be able to do these as github suggestions on PRs, which would be low enough friction that I'd support moving it into presubmit, but I think the contributor experience is more important here.

/approve

PR #406, opened 2019-05-10
awssqs: fix breakage after moving to contrib/

srvaroa:
Fixes two issues introduced at f62c2387. (See https://github.com/knative/eventing-sources/pull/253)

### DIFF
--- a/contrib/awssqs/config/201-clusterrole.yaml
+++ b/contrib/awssqs/config/201-clusterrole.yaml
@@ -22,13 +22,27 @@ rules:
   resources:
   - deployments
   verbs:
-  - *
+  - get
+  - list
+  - watch
+  - create
+  - update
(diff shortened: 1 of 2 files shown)


srvaroa:
NOTE: I have not yet verified this on a running cluster.

nachocano:
great!!! Thanks @srvaroa 
/lgtm

srvaroa:
Confirmed to be working with eventing-sources @0451142c and eventing @53ee52c1 

Note that it'll fail with v0.5 due to "no matches for kind "EventType" in version "eventing.knative.dev/v1alpha1" as it needs the dependency on eventing v0.6.

nachocano:
/assign @n3wscott
This one also needs to be merged before the 0.6 cut. Thanks!

PR #253, opened 2019-03-16
Move AWS SQS source to contrib

srvaroa:
Along with some documentation cleanup and a couple of simplifications of code.  Tested on v0.4.0.

PR #408, opened 2019-05-10
AWS is out in contrib, hence removing

matzew:
Follow up for #253 

@srvaroa @n3wscott


srvaroa:
Thanks! Sorry I missed this one

matzew:
@n3wscott @nachocano this should land before the 0.6.0 is cut

PR #413, opened 2019-05-11
allow controller to get/list Services as sinks

iancoffey:
## Proposed Changes

Now that Services can be used as Sinks, the controller needs the ability to Get/List/etc them or it fails.

Related to: https://github.com/knative/eventing-sources/pull/412


iancoffey:
This one could use a review if anyone has the time

PR #415, opened 2019-05-13
Reconcile EventType upon Source change

nachocano:
## Proposed Changes

- Bug found while editing the Event Sources. E.g., changing its broker, or the topic, etc.
- Properly reconciling the EventTypes upon Event Source change, and also upon EventType change.
- Updating eventing version to have an EventType fix from the spec.description field. Thanks @n3wscott.
- Pending: if the Source sink changes to a non-Broker one, we are not properly removing the EventTypes. Will do in a follow up PR.

**Release Note**

```release-note
eventing-sources 0.6 needs eventing 0.6. If you are updating to the new sources, then also update eventing to v0.6.
```


nachocano:
/assign @n3wscott 
Similar bug in Eventing Sources. Need to properly update the types in the registry upon changes in Sources. I still need a follow up PR when changing to non-broker sinks, but this should go in asap before the 0.6 cut.

PR #424, opened 2019-05-15
GCPPubSub Transformer cosmetic changes

nachocano:
## Proposed Changes

  * Missing changes from Adam's comments in previous [PR](https://github.com/knative/eventing-sources/pull/398)
  * Removing ce-type override. That should be done in a specific transformer.
    
**Release Note**

```release-note
CloudEvents changes for GCPPubSubSource should be done in transformers.
```


contrib/gcppubsub/config/300-gcppubsubsource.yaml
```diff
@@ -48,7 +48,7 @@ spec:
               description: "Credential to use to poll the GCP PubSub Subscription. It is not used to create or delete the Subscription, only to poll it. The value of the secret entry must be a service account key in the JSON format (see https://cloud.google.com/iam/docs/creating-managing-service-account-keys)."
             googleCloudProject:
               type: string
-              description: "ID of the Google Cloud Project that the PubSub Topic exists in."
+              description: "ID of the Google Cloud Project that the PubSub Topic exists in. E.g. 'my-project-1234' rather than its display name, 'My Project' or its number '1234567890'."
             serviceAccountName:
               type: string
               description: "Name of the ServiceAccount that will be used to run the Receive Adapter Deployment."
```
nachocano: re: Adam's comment "Give an example. Is this the full email address? Just the name before the '@'?"

I'm not sure, it was just a copy/paste from the types.go comment. Is it a full email address?

contrib/gcppubsub/cmd/receive_adapter/main.go
```diff
@@ -61,7 +61,7 @@ func main() {
 
 	var env envConfig
 	if err := envconfig.Process("", &env); err != nil {
-		log.Fatal("Failed to process env var", zap.Error(err))
+		logger.Fatal("Failed to process env var", zap.Error(err))
```
Harwayne: Is this a sugared logger? If so, it should be `Fatalw` (my preference is just switch to a desugared logger to make sure these bugs don't occur).

contrib/gcppubsub/config/300-gcppubsubsource.yaml
Harwayne: My mistake on the earlier comment. This field is about the K8s Service Account, not any GCP service account. How about:

```
Name of the Kubernetes ServiceAccount that is running the Receiver Adapter Pods.
```

contrib/gcppubsub/cmd/receive_adapter/main.go
nachocano: I don't think it's a sugared one, as I cannot Desugar it. It has the Sugar() method, which makes me think is not sugared?

`logger, err := logCfg.Build()` creates a sugared one?

contrib/gcppubsub/config/300-gcppubsubsource.yaml
nachocano: Done!

contrib/gcppubsub/cmd/receive_adapter/main.go
Harwayne: Ah, my mistake. I read `log.Fatalf` above and thought it was the same logger.

PR #382, opened 2019-04-30
add gitlab source to contrib

vincent-pli:
## Proposed Changes
Add gitlab source to `contrib/`

Add new crd: GitLabSource and related Controller

PR #450, opened 2019-05-31
[WIP] gitlabsource

sebgoa:
This is a potential move of https://gitlab.com/triggermesh/gitlab-source which is already listed in https://github.com/knative/docs/tree/master/docs/eventing/sources

To be compared with #382 and discuss what goes in `/contrib`


sebgoa:
Fwiw, I don't think we should have a `/contrib` this will become a kitchen sink like `kubernetes/contrib` became. It tooks years to clean up `kubernetes/contrib`

If we want to have sources in the knative org we should put them in their own repos, we don't need a mono repo of sources.

evankanderson:
Suggestion to add OWNERS file and OWNERS_ALIASES entries ala Kafka contrib/ location:

https://github.com/knative/eventing-sources/blob/master/contrib/kafka/OWNERS

https://github.com/knative/eventing-sources/blob/master/OWNERS_ALIASES

matzew:
what's the state here, compared to https://github.com/knative/eventing-contrib/pull/382 ?

sebgoa:
This is more of a "governance/policy" issue.

We have a list of sources defined in our docs, they point to internal (within the knative repo) and external sources.

The question is whether the community endorses one specific source and that this source needs to be in the knative org ?

I am happy to contribute all the Triggermesh sources back to the knative repo, but early discussion hinted at keeping sources external.

n3wscott:
It is valid to have an external source that is knative compatible. There is nothing that says it must be owned and inside of the Knative repo.

sebgoa:
of course.

But as soon as there is a source that lands in the knative repo it will be perceived as the reference source for that "kind". Even if there is a pre-existing source listed in the official knative listing.

Does that make sense ?

matzew:
> Does that make sense ?

@sebgoa it does - want to update this, and we get it in ? 

I guess this is like... https://gitlab.com/triggermesh/gitlabsource ?

sebgoa:
ok, we will update and get it in

matzew:
> ok, we will update and get it in

@sebgoa any updates on this ?

sebgoa:
asking @tzununbekov to look at it.

sebgoa:
@matzew closing this in favor of #952

PR #459, opened 2019-06-11
Fix #458: upgrade to camel k 0.3.4

### LINKED ISSUE #458: Upgrade camel sources to use latest camel k
Camel K 0.3.4 released.

https://github.com/apache/camel-k/releases/tag/0.3.4

nicolaferraro:
Fixes #458

This upgrades camel sources to use latest camel k.


contrib/camel/config/300-camelsource.yaml
```diff
@@ -65,54 +65,6 @@ spec:
                       type: string
                   type: object
                 integration:
-                  properties:
```
evankanderson: Why are these properties removed?

nicolaferraro: Users are facing issues with using different versions of Camel K. Those properties come from [here](https://github.com/apache/camel-k/blob/master/assets/json-schema/Integration.json).

Currently the Knative camel sources are strictly bound to a specific version of camel k. In fact switching to even a different minor version of camel k sometimes breaks everything. E.g. when from camel k side we add a field to the spec, unmarshalling fails in knative-camel-controller because of the unknown field.

So, long term I'd like to support multiple camel k releases in knative camel sources to avoid such problems and this means we cannot set a fixed schema here.

This is just a first step, I preferred dropping it instead of upgrading it to the camel k 0.3.4 version.

evankanderson: Great, I wanted to make sure this wasn't a broken merge.

nicolaferraro:
There were some deps added by `dep ensure` that now I've removed using `update-deps.sh`. It should be ok now.

PR #460, opened 2019-06-11
Fix #457: remove limits from camel controller manager

### LINKED ISSUE #457: Camel controller manager sometimes does not start
## Expected Behavior

It always starts.

## Actual Behavior

Sometimes it remains in "ContainerCreationError" on Openshift. Removing pod limits seems to solve the issue.

  1. Install Eventing and Serving
  2. Install Camel K cluster resources
  3. Install Camel sources

## Additional Info

@matzew I think this happens also to Kafka contrib.
I think it would be good to remove resource limitations.

nicolaferraro:
Fixes #457

Removing limits prevent the controller to crash sporadically on startup. Increasing them would be another option, but the controller does not require so many resources.

### DIFF
--- a/contrib/camel/config/500-controller.yaml
+++ b/contrib/camel/config/500-controller.yaml
@@ -32,5 +32,9 @@ spec:
       containers:
       - image: github.com/knative/eventing-contrib/contrib/camel/cmd/controller
         name: manager
+        resources:
+          requests:
+            cpu: 20m
+            memory: 20Mi
       serviceAccount: camel-controller-manager
       terminationGracePeriodSeconds: 10
--- a/contrib/camel/config/500-controller.yaml
+++ b/contrib/camel/config/500-controller.yaml
@@ -32,12 +32,5 @@ spec:
       containers:
       - image: github.com/knative/eventing-contrib/contrib/camel/cmd/controller
         name: manager
-        resources:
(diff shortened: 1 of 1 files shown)


contrib/camel/config/500-controller.yaml
```diff
@@ -32,12 +32,5 @@ spec:
       containers:
       - image: github.com/knative/eventing-contrib/contrib/camel/cmd/controller
         name: manager
-        resources:
-          limits:
-            cpu: 100m
-            memory: 30Mi
-          requests:
-            cpu: 20m
-            memory: 20Mi
```
evankanderson: You might want to leave `requests` (which is a minimum guarantee), but remove `limits` (which is a hard maximum).

nicolaferraro: Thanks @evankanderson, it seems to work well with only requests. I've added them back :+1:

duglin:
Can we just increase the limits instead of removing them? W/o limits the pods ca become noisy neighbors
@nicolaferraro

matzew:
I'd think that if we add those, we need to have confidence in the actual values - I don't see value in blind guesses, so to have some, just to have some ...  See: https://bugzilla.redhat.com/show_bug.cgi?id=1714183

PR #462, opened 2019-06-12
Removing limits on Kafka controller

### LINKED ISSUE #457: Camel controller manager sometimes does not start

matzew:
Fixes #457 (for Kafka).  Thanks @nicolaferraro for filing the (generic) issue

### DIFF
--- a/contrib/kafka/config/500-controller.yaml
+++ b/contrib/kafka/config/500-controller.yaml
@@ -37,9 +37,6 @@ spec:
           value: github.com/knative/eventing-contrib/contrib/kafka/cmd/receive_adapter
         volumeMounts:
         resources:
-          limits:
-            cpu: 100m
-            memory: 30Mi
           requests:
             cpu: 20m
             memory: 20Mi


duglin:
Why are we allowing pods to each up as much resources as they want?

matzew:
I'd think that if we add those, we need to have confidence in the actual values - I don't see value in blind guesses, so to have some, just to have some ...  See: https://bugzilla.redhat.com/show_bug.cgi?id=1714183

duglin:
agreed to trying to pick smarter values, so let's grab the values used to fix that issue instead of just removing them

evankanderson:
Limits (and requests) may also need to be dynamic based on cluster size/number of events in flight. [Vertical Pod Autoscaling](https://github.com/kubernetes/autoscaler/tree/master/vertical-pod-autoscaler) might be one way to set those limits dynamically.

alanconway:
What is it we are trying to limit?

a) Ensure enough resources to get to some "standard" steady state of operation.
b) Ensure enough resources to maintain some desired scale/size/load.
c) Ensure no more resources than the system designer is willing to give it.

We can do a) in theory and it would be great if we provided scaling guides for people trying to figure out b) and c), but it's not trivial. We need to understand and measure relationship between scale, load and resources in our code and in our dependencies (Kafka libs etc.) I'm not saying we shouldn't, but it is extra documentation, testing and benchmarking, and it will impact the pace of innovation.

PR #476, opened 2019-06-27
Unit test for status.address.hostname

### LINKED ISSUE #475: GitHubSource doesn't work with Serving 0.7
## Expected Behavior
I can target a Serving 0.7 Service with a GitHubSource.

## Actual Behavior
Doesn't work because Service 0.7 doesn't populate the `status.address.hostname` field.

## Additional Info
Probably an issue for all sources.
See https://knative.slack.com/archives/C9JP909F0/p1561666023287200

Harwayne:
## Proposed Changes

  * Add a unit test for #476.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controller/sinks/sinks_test.go
+++ b/pkg/controller/sinks/sinks_test.go
@@ -32,6 +32,7 @@ import (
 
 var (
 	addressableDNS = "addressable.sink.svc.cluster.local"
+	addressableURL = fmt.Sprintf("http://%s/", addressableDNS)
 
 	addressableName       = "testsink"
 	addressableKind       = "Sink"
@@ -52,7 +53,7 @@ var (
 
 func init() {
 	// Add types to scheme
-	duckv1alpha1.AddToScheme(scheme.Scheme)
+	_ = duckv1alpha1.AddToScheme(scheme.Scheme)
 }
 
 func TestGetSinkURI(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/controller/sinks/sinks_test.go
```diff
@@ -32,6 +32,7 @@ import (
 
 var (
 	addressableDNS = "addressable.sink.svc.cluster.local"
+	addressableURL = fmt.Sprintf("http://%s/", addressableDNS)
```
matzew: In #477 I was doing a _straight_ port of what @mattmoor did,  so, only updating the 

```
			want:      fmt.Sprintf("http://%s", addressableDNS),
```

Note, that the format here is different `%s` versus `%s/` (trailing `/`) on your PR

matzew:
generally wondering, what best to avoid some of this code duplications, here and eventing (core)

matzew:
@Harwayne in 477, I had to (similar to Matt) also fix some of the URI expectations, on the tests

n3wscott:
In v1beta1 GetURL goes away.

pkg/controller/sinks/sinks_test.go
grantr: Agree with @matzew, we should probably not add a trailing slash here. Golang's `url.URL.String()` doesn't: https://play.golang.org/p/cFaC63TxyXd

pkg/controller/sinks/sinks_test.go
```diff
@@ -63,7 +64,15 @@ func TestGetSinkURI(t *testing.T) {
 		wantErr   error
 		ref       *corev1.ObjectReference
 	}{
-		"happy": {
+		"happy - hostname": {
```
grantr: I like the addition of this test ensuring we can still target v1alpha1 Addressables.

pkg/controller/sinks/sinks_test.go
Harwayne: Removed.

PR #479, opened 2019-07-03
Make gcloud a no-op if not available for test scripts

alanconway:
Fixes the same issue as knative/eventing#1507
## Proposed Changes

If gcloud is unavailable, replace it with a no-op shell function

### DIFF
--- a/test/upload-test-images.sh
+++ b/test/upload-test-images.sh
@@ -14,6 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# If gcloud is not available make it a no-op, not an error.
+which gcloud &> /dev/null || gcloud() { echo "[ignore-gcloud $*]" 1>&2; }
+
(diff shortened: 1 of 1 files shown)


alanconway:
On Mon, Jul 8, 2019 at 11:29 AM Ying Chun Guo <notifications@github.com>
wrote:

> *@daisy-ycguo* commented on this pull request.
>
> Actually, the current integrated CI jobs are just based on gcloud. So the
> CI testing doesn't make sense for this PR. We should test it in an
> environment without gcloud.
>

That would be great. I hit this problem running the e2e tests on a homespun
kubeadm cluster and on minikube. The only script I currently depend on
is upload-test-images.sh, but it would be nice to get all the test scripts
(esp. the pre-submit script) to run cleanly on non-gcloud environments.
Since I don't use gcloud I'm wary of messing with this too much myself.

PR #496, opened 2019-07-31
Use knative.dev/test-infra

chaodaiG:


grantr:
@chaodaiG This PR used an older version of dep that made unnecessary changes to Gopkg.lock. Can you run `hack/update-deps.sh` with the new version of dep and create a new PR so we can revert the unnecessary changes? Thanks!

chaodaiG:
sure, done

PR #498, opened 2019-07-31
verify codegen also check Gopkg.lock

chaodaiG:
Fixes https://github.com/knative/test-infra/issues/1213


chaodaiG:
I don't see any reason why not, this file is always copy pasted among repos afaik

grantr:
If it's not too much trouble, can you update this PR with the copy from eventing? That way they'll at least be identical from here forward. :smiley_cat:

chaodaiG:
done :)

PR #499, opened 2019-08-01
Add kafka channels

matzew:
## Proposed Changes

  * add channels for apache kafka (CRD/CCP)
  *
  *

**Release Note**

```release-note
```


matzew:
@vaikas-google I tested this w/ two samples:

* source to channel (CRD)
* source to channel (CCP)

See here:  https://github.com/matzew/knative-eventing-samples/pull/1

matzew:
IMO once this is in, we can also merge ->  https://github.com/knative/eventing/pull/1609

Gopkg.toml
```diff
@@ -71,7 +81,7 @@ required = [
 [[constraint]]
   name = "github.com/knative/eventing"
   # HEAD as of 2019-06-26. Needed to import EventType API object.
```
vaikas-google: is this right? seems like revision changed, but not the date?

kafka/channel/config/README.md
```diff
@@ -0,0 +1,92 @@
+# Apache Kafka Channels
+
+Kafka channels are those backed by [Apache Kafka](http://kafka.apache.org/)
+topics.
+
+## Deployment steps
+
+1. Setup [Knative Eventing](../../../DEVELOPMENT.md)
```
vaikas-google: This points to setting up the dev stack (including things like ko, which is probably not necessary just to install kafka channels?)

vaikas-google: nm, I see ko being used below. However, we should probably have a kubectl version and released binaries for people so that they can just install a release version.

kafka/channel/pkg/controller/channel/provider_test.go
```diff
@@ -0,0 +1,40 @@
+/*
```
vaikas-google: should we update all these dates to 2019?

kafka/channel/config/README.md
matzew: sure, let me do that in a different PR - this was mainly just porting it over and test it

kafka/channel/pkg/controller/channel/provider_test.go
matzew: ok, can do that in a different PR

vaikas-google:
These are moving over from the eventing repo and hence the coverage issues are not new and don't want to hold off merging this based on that.

PR #451, opened 2019-05-31
[WIP] IBM MQ source

sebgoa:
This is a proposed new container source, it gets events from IBM MQ queues.

It is self-contained and moved as is in `/contrib`

It does not use `ko` yet.

PR #500, opened 2019-08-02
IBM MQ containersource

tzununbekov:
Containersource that reads events from IBM MQ
Fixes previous PR #451 

Requires CGO module to work with IBM bindings, overrides ko's base image, ships mq lib.  
Can be deployed with `ko` (from repository root): 
```
CGO_ENABLED=1 ko apply -f contrib/ibm-mq/
```


contrib/ibm-mq/main.go
```diff
@@ -0,0 +1,183 @@
...
+
+	cloudEventsClient := cloudevents.NewClient(
+		sink,
+		cloudevents.Builder{
+			Source:    "ibm:mq",
+			EventType: "message queue item",
+		},
+	)
+
+	fmt.Printf("%+v\n", cfg)
```
lionelvillard: use logger.

contrib/ibm-mq/main.go
```diff
@@ -0,0 +1,183 @@
...
+	// Opening a QUEUE (rather than a Topic or other object type) and give the name
+	mqod.ObjectType = ibmmq.MQOT_Q
+	mqod.ObjectName = cfg.QueueName
+
+	qObject, err := qMgrObject.Open(mqod, openOptions)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Println("Opened queue", qObject.Name)
```
lionelvillard: same here

tzununbekov: thanks @lionelvillard, I've entirely got rid of fmt methods in the app

contrib/ibm-mq/README.md
```diff
@@ -0,0 +1,116 @@
+## IBM MQ source
+
+This event source is meant to be used as a Container Source within a Knative cluster in order to consume messages from IBM Message Queue.
```
duglin: nit: is it worth mentioning the implementation detail that it's written as a "container source"? The user of this event source doesn't need to know, or care, do they?

contrib/ibm-mq/README.md
```diff
@@ -0,0 +1,116 @@
...
+    "message_data": "Hello Knative"
+  }
+```
+
+### Knative usage
+
+Edit the Container source manifest to specify your queue and connection parameters then apply it like so:
+
+```
+kubectl apply -f containersource.yaml
```
duglin: any reason it's not named `mqsource.yaml` ?

contrib/ibm-mq/containersource.yaml
```diff
@@ -0,0 +1,30 @@
...
+              value: mq-service.default.svc.cluster.local(1414)
+            - name: QUEUE_NAME
+              value: DEV.QUEUE.1
+            - name: USER_ID
+              value: <USER_ID>
+            - name: PASSWORD
+              value: <PASSWORD>
+  sink:
+    apiVersion: eventing.knative.dev/v1alpha1
+    kind: Channel
```
duglin: Given the switch to brokers/triggers, should this be "Broker" instead?  I'm still new to the broker/channel split so it might be a silly question.

contrib/ibm-mq/main.go
```diff
@@ -0,0 +1,178 @@
...
+	connOptions.ClientConn = channelDefinition
+	connOptions.SecurityParms = connSecParams
+
+	// And now we can try to connect. Wait a short time before disconnecting.
+	qMgrObject, err := ibmmq.Connx(cfg.QueueManager, connOptions)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Printf("Connection to %s succeeded.\n", cfg.QueueManager)
```
duglin: nit: use `%q` so it's quoted. If the name happens to sound too English then it'll read like a normal sentence and not like the name of a queue.

duglin: oh never mind - not too important, this is just a log msg - I thought it was printed for the end user.

contrib/ibm-mq/main.go
```diff
@@ -0,0 +1,178 @@
...
+
+		wg.Add(1)
+		go func(md *ibmmq.MQMD, messageData []byte) {
+			defer wg.Done()
+			data := string(bytes.Trim(messageData, "\u0000"))
+			msg := MQMessage{
+				Message:     md,
+				MessageData: data,
+			}
+			log.Printf("New message: %s", data)
```
duglin: two things:
1 - is it safe from a security perspective to log the message contents itself? seems risky
2 - since there should be a ton of these running in parallel, I wonder if it would be helpful to be able to correlate all log messages from within a single thread? Meaning, prefix the logs with some kind of unique ID per thread. Then someone can correlate the error on line 156 with the incoming message. Of course, if you kill the above log then this isn't needed  :-)

duglin:
overall it looks good to me - just a couple of minor comments

contrib/ibm-mq/README.md
tzununbekov: Agree, removed "container source" mention

contrib/ibm-mq/README.md
tzununbekov: sure, renamed manifest to `mqsource.yaml`

contrib/ibm-mq/containersource.yaml
tzununbekov: Not sure about that. Changing type to `Broker` won't be enough to make it work as broker I guess.

contrib/ibm-mq/main.go
tzununbekov: Also fixed for better logging

contrib/ibm-mq/main.go
tzununbekov: Removed message logging - it was there for debug purpose ;)

contrib/ibm-mq/containersource.yaml
duglin: Perhaps @n3wscott can comment on this?  Scott - which would make most sense, Broker or Channel?

tzununbekov:
thanks for the review @duglin but there is an issue with this event source - IBM MQ go client [requires](https://github.com/ibm-messaging/mq-golang#using-the-package) MQ itself to be installed on the machine where build happens. I don't see how I can make it pass build on CI, even though it's not needed to be ko apply-able.

evankanderson:
@adrcunha for thoughts on how to handle this requirement for `cgo` builds for this message source.

Separately, I'll point out that Google is running separate CI and releases for Google-specific integrations (PubSub, Cloud Storage, etc) at: https://github.com/google/knative-gcp and TriggerMesh is doing the same at https://github.com/triggermesh/knative-lambda-sources, etc, so if the CI here is being difficult, we want to make cross-linking to vendor repos easy and first-class on https://knative.dev/.

adrcunha:
You should be able to install IBM MQ in the E2E setup phase. Use the instructions at

https://github.com/ibm-messaging/mq-golang#common

and hook them to a `cluster_setup()` function in `e2e-tests.sh` like in

https://github.com/knative/client/blob/dbb63d45429a102ef499586069c8bf21a88f9f7c/test/e2e-tests.sh#L36

tzununbekov:
> You should be able to install IBM MQ in the E2E setup phase. Use the instructions at
> 
> https://github.com/ibm-messaging/mq-golang#common
> 
> and hook them to a `cluster_setup()` function in `e2e-tests.sh` like in
> 
> https://github.com/knative/client/blob/dbb63d45429a102ef499586069c8bf21a88f9f7c/test/e2e-tests.sh#L36

looks like `e2e-tests.sh` being executed only for integration tests ([here](https://github.com/knative/test-infra/blob/master/scripts/presubmit-tests.sh#L261)), so I had to put required env variables in `presubmit-tests.sh` script:
https://github.com/knative/eventing-contrib/pull/500/files#diff-8f99251cc2fba45d21cb371ad880c91c

MQ dependencies are stored in kodata so they could be used in both, local build and for "ko apply".

@evankanderson @adrcunha could you take a look at this once again please?

adrcunha:
Indeed e2e-tests.sh only runs for integration tests. I think something is missing here, because now there are only a few env vars added to presubmit-tests.sh. Are they required because of the new code in contrib?

tzununbekov:
> Indeed e2e-tests.sh only runs for integration tests. I think something is missing here, because now there are only a few env vars added to presubmit-tests.sh. Are they required because of the new code in contrib?

right, these env variables are the only thing required to build MQ source - they are pointing compiler to the `kodata` directory where MQ libs are stored

sebgoa:
@adrcunha could you check this again, see if @tzununbekov works for you ?
@evankanderson any suggestions on how to get this moving ?

duglin:
@tzununbekov the IBM MQ team has started to review this, but due to time constraints their comments may not land until next week.  Just wanted you to know that it's on their radar ;-)

tzununbekov:
@duglin thanks for the heads up!

matrober-uk:
Hi @tzununbekov - thanks for submitting this - on behalf of the IBM MQ development team we're pleased to see you submit a KNative event source for IBM MQ, and we're looking forward to seeing how customers will use it!

We have reviewed within the MQ development team and would like to share the following review comments/requests please.

Thanks again for your contribution!

Comments:
- Suitability for production
  - There are a few changes that can be made to harden the code to make it more suitable for production style usage
  - In particular we would recommend adding appropriate error handling so that messages can be retried if the delivery to the event client fails
  - This could/should include consuming the message under a sync point (transaction) so that it can be rolled back and redelivery attempted by the MQ infrastructure
  - https://github.com/knative/eventing-contrib/pull/500/files#diff-401655c4da416f7172aea1639245b63bR153
  - Similarly, what is the behaviour if the connection to the queue manager is broken while the loop is running (e.g. queue manager is restarted) - the GET will fail with an error and break out of the for loop - does the main method automatically get re-run from the top by the framework in that scenario in order to establish a new connection to the queue manager?
  - What sort of retry behaviour is there in the framework that controls the main method - for example if the queue manager was unavailable at the time that main.go is executed? 

- Licensing of the IBM MQ redistributable client files
  - There are a number of IBM MQ files (.h and .so) checked in to /contrib/ibm-mq/kodata/inc which are licensed under the IPLA license and not the Apache 2.0 license that covers the repo as a whole
  - The IPLA section on "Redistributables" also has a series of conditions which are potentially violated by checking in copies of the files, in particular clause 1 (object code only) and probably also 3, 11-ii and 11-iii
  - To avoid these licensing issues these files should be extracted dynamically from the redist client download as part of the build process, and removed from the checked in source tree please

- If possible, please update the redist client import (https://github.com/knative/eventing-contrib/pull/500/files#diff-c449959a02664b1bdb78228113d3bcd2R5) to the LTS 9.1.0.3 package
  - Unless there are specific functions in the 9.1.x CD releases that is required?
  - (for example ccdt_schema.json is not included in the v9.1.0.3 download - is that file required for the runtime?)
  - (ccsid.tbl, ccdid_part2.tbl are included in v9.1.0.3, as are the .so files)
  - Using the LTS release will ensure a long support window for this library, rather than requiring it to be updated every 12 months in order to stay in support
  - If you can't switch to the LTS option then please update to the latest MQ CD release v9.1.3 to give the maximum possible window for support (July 2020 for v9.1.3)

- Please also update the mq-golang dependency version to 4.1.0 in order to keep up to date with the latest there;
  - https://github.com/knative/eventing-contrib/pull/500/files#diff-db5d2e36b7f6907ca4bd4d2d2e45a263R30

- Please update the phrase "IBM Message Queue" to either "IBM MQ" or "an IBM MQ queue" (the product name MQ is never expanded)
  - https://github.com/knative/eventing-contrib/pull/500/files#diff-2adb19385e6b3a9e58c8f38e632898e3R3

- For the WaitInterval of 3 seconds, since this code is running in an infinite loop is there any reason that this should be made slightly longer, such as 10 or 15 seconds?
  - https://github.com/knative/eventing-contrib/pull/500/files#diff-401655c4da416f7172aea1639245b63bR128
  - This would reduce the churn on the Go process

- From a Go style perspective, can we suggest not checking in the /vendor directory, and instead allowing it to be generated at build time?

- We suggest removing the commented out line that tries to set the ApplName
  - https://github.com/knative/eventing-contrib/pull/500/files#diff-401655c4da416f7172aea1639245b63bR82
  - Note that this is one aspect which will likely not be supported by the 9.1.0.3 redistributable client, but on balance it is better to have the longer support window

evankanderson:
Thanks @matrober-uk for the review of the IBM MQ code; I'm not familiar with 

I'm taking a look at this right now, and I have a couple concerns:

1. Why is this in `contrib/ibm-mq`? All the other components are in top-level directories.
2. Why do you have a separate `Gopkg.toml` and `Gopkg.lock`? None of the other contrib components have separate dependencies like this.
3. I'm uncomfortable with checking in the `.so` library files without (at a minimum) some README indicating:
   1. Where they came from
   2. How to refresh them
   3. Any licensing requirements
4. @sebgoa -- can you confirm that you will be on the hook for fixes to this component going forward? If not, @tzununbekov or someone else will need to be on the hook for this.

Comments on individual sections of code will follow in a separate review. Some emphasis on @matrober-uk 's comments.

> Hi @tzununbekov - thanks for submitting this - on behalf of the IBM MQ development team we're pleased to see you submit a KNative event source for IBM MQ, and we're looking forward to seeing how customers will use it!
> 
> We have reviewed within the MQ development team and would like to share the following review comments/requests please.
> 
> Thanks again for your contribution!
> 
> Comments:
> 
> * Suitability for production
>   
>   * There are a few changes that can be made to harden the code to make it more suitable for production style usage
>   * In particular we would recommend adding appropriate error handling so that messages can be retried if the delivery to the event client fails
>   * This could/should include consuming the message under a sync point (transaction) so that it can be rolled back and redelivery attempted by the MQ infrastructure
>   * https://github.com/knative/eventing-contrib/pull/500/files#diff-401655c4da416f7172aea1639245b63bR153

+1 to this suggestion; the expectation is that 

>   * Similarly, what is the behaviour if the connection to the queue manager is broken while the loop is running (e.g. queue manager is restarted) - the GET will fail with an error and break out of the for loop - does the main method automatically get re-run from the top by the framework in that scenario in order to establish a new connection to the queue manager?

If the container exits, a new container will be started. Restarting the loop seems like it could be an incremental improvement.

>   * What sort of retry behaviour is there in the framework that controls the main method - for example if the queue manager was unavailable at the time that main.go is executed?

The container source Deployment will keep launching new Pods when the previous Pod dies. This will generate some amount of extra churn in the cluster, but is probably not dire.

> * Licensing of the IBM MQ redistributable client files
>   
>   * There are a number of IBM MQ files (.h and .so) checked in to /contrib/ibm-mq/kodata/inc which are licensed under the IPLA license and not the Apache 2.0 license that covers the repo as a whole

Is the IPLA in question this license: https://www-03.ibm.com/software/sla/sladb.nsf/pdf/ipla/$file/ipla_en.pdf ?

Alternatively, is it the IPL: https://opensource.org/licenses/IPL-1.0 ?

At a minimum, there needs to be a LICENSE or LICENSE.txt file **in text format** alongside the code. Since this is native code, it's not clear that https://github.com/knative/test-infra/tree/master/tools/dep-collector will find it, so additional work may need to be done to get license generation working correctly.

Once we know which license this is, we may also need Google's legal team to sign off on the IPL usage here. I'm guessing IPL is probably okay, but **I am not a lawyer**.

>   * The IPLA section on "Redistributables" also has a series of conditions which are potentially violated by checking in copies of the files, in particular clause 1 (object code only) and probably also 3, 11-ii and 11-iii
>   * To avoid these licensing issues these files should be extracted dynamically from the redist client download as part of the build process, and removed from the checked in source tree please
> * If possible, please update the redist client import (https://github.com/knative/eventing-contrib/pull/500/files#diff-c449959a02664b1bdb78228113d3bcd2R5) to the LTS 9.1.0.3 package

How do these licensing terms apply to redistributing a docker container containing the combination of IPLA software and Knative AGPL software? Is this a case where we'd need to ask each customer to build that combination themselves?

>   * Unless there are specific functions in the 9.1.x CD releases that is required?
>   * (for example ccdt_schema.json is not included in the v9.1.0.3 download - is that file required for the runtime?)
>   * (ccsid.tbl, ccdid_part2.tbl are included in v9.1.0.3, as are the .so files)
>   * Using the LTS release will ensure a long support window for this library, rather than requiring it to be updated every 12 months in order to stay in support
>   * If you can't switch to the LTS option then please update to the latest MQ CD release v9.1.3 to give the maximum possible window for support (July 2020 for v9.1.3)
> * Please also update the mq-golang dependency version to 4.1.0 in order to keep up to date with the latest there;
>   
>   * https://github.com/knative/eventing-contrib/pull/500/files#diff-db5d2e36b7f6907ca4bd4d2d2e45a263R30

`Gopkg.lock` seems to indicate version = "v4.1.0"

> * Please update the phrase "IBM Message Queue" to either "IBM MQ" or "an IBM MQ queue" (the product name MQ is never expanded)
>   
>   * https://github.com/knative/eventing-contrib/pull/500/files#diff-2adb19385e6b3a9e58c8f38e632898e3R3
> * For the WaitInterval of 3 seconds, since this code is running in an infinite loop is there any reason that this should be made slightly longer, such as 10 or 15 seconds?
>   
>   * https://github.com/knative/eventing-contrib/pull/500/files#diff-401655c4da416f7172aea1639245b63bR128
>   * This would reduce the churn on the Go process
> * From a Go style perspective, can we suggest not checking in the /vendor directory, and instead allowing it to be generated at build time?

The `dep` tool automatically vendors the Go code. Additionally, we've found that fetching dependencies dynamically introduces a wide range of failure modes around availability of the original package (including cases like package deletion or repo unavailability).

We've decided to prefer duplicating the bytes in preference to the availability risks of late-binding.

> * We suggest removing the commented out line that tries to set the ApplName
>   
>   * https://github.com/knative/eventing-contrib/pull/500/files#diff-401655c4da416f7172aea1639245b63bR82
>   * Note that this is one aspect which will likely not be supported by the 9.1.0.3 redistributable client, but on balance it is better to have the longer support window

sebgoa:
@evankanderson yes we will be on the hook, but if you don't want this in the repo please say so now and we will maintain it outside.

It is not clear to me what is the future of this repo since the google knative source is now outside of it.

evankanderson:
The most critical item is clarification of the license status of this code vis-a-vis redistribution. Beyond that, I think we intend to keep maintaining the contributed items here, and adding them where they make sense (mostly consumers for OSS systems like Kafka and NATSS, but also common infrastructure which may not provide its own importer, like Amazon and GitHub).

I'm fine with having the IBM importer here if the packaging/licensing can be aligned with the rest of the items here.

contrib/ibm-mq/main.go
```diff
@@ -0,0 +1,175 @@
...
+	"log"
+	"sync"
+
+	"github.com/caarlos0/env"
+	"github.com/ibm-messaging/mq-golang/ibmmq"
+	"knative.dev/pkg/cloudevents"
+)
+
+var sink string
+var wg sync.WaitGroup
```
evankanderson: Why declare this out here, rather than in main?

evankanderson:
FWIW, it may make more sense to see whether the test-infra tooling can work with a Dockerfile for parts of the repo, though it's unfortunate that we'll lose capabilities like [dep-collector](https://github.com/knative/test-infra/tree/master/tools/dep-collector) to automatically manage the LICENSE files shipped with each Docker container.

tzununbekov:
@matrober-uk @evankanderson thanks for the review and comments, I'm working on incorporating changes you suggested. In general I see three directions for me here:

1. message delivery reliability - ensure "at-least-once" delivery approach, improve code testability
1. remove `.h` and `.so` files from the repository - figure out how to dynamically load redist libs during the build (though there is no actually scripted build step). Make sure to use latest LTS versions of MQ dependencies
1. licensing - I'm not sure where we are in this question. Does it make sense now to work on these changes, if at the end it turns out that the code cannot be merged due to a license conflict?

I'll gradually update the code with these changes while the licensing question is something that I'd like to clarify as soon as possible. Thanks again!

test/presubmit-tests.sh
```diff
@@ -25,6 +25,12 @@
 # in a net-negative contributor experience.
 export DISABLE_MD_LINTING=1
 
+export CGO_ENABLED=1
```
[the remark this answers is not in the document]
tzununbekov: right, I couldn't find a way to enable CGO for a single test. I will be glad to any advice :)

contrib/ibm-mq/Makefile
```diff
@@ -0,0 +1,15 @@
+TARGET=mq
```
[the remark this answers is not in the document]
tzununbekov: Makefile and Dockerfile remained after our development process in local test env. I'll remove them in favor of `ko`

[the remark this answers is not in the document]
evankanderson: `ko` might not be the right tool here, if you need to use CGO and extract files from the redistributable object (due to licensing requirements).

Ideally, this would be handled by `dep` when you depend on the `mq-golang` package, but that may not be possible. In that case, figuring out how to release this using a Dockerfile may be the path of least resistance.

test/presubmit-tests.sh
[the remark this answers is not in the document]
evankanderson: How about [using build tags](https://golang.org/pkg/go/build/) (suggested by mattmoor on Slack):

In your `_test.go` files, near the top:

```golang
// +build cgo
```

And then change `test/presubmit-tests.sh` to run as:

```shell

source $(dirname $0)/../vendor/knative.dev/test-infra/scripts/presubmit-tests.sh

# Run the default build, unit and integration tests without cgo

main $@

# Re-run the tests using cgo for the ibm-mq source. Avoid contaminating the environment by using a sub-shell.
(
export CGO_ENABLED=1
export ...

# TODO: figure out how to pass `-tags cgo` to `go test` here to only run cgo tagged tests
main $@
)

```

sebgoa:
Thank you all for the reviews, it seems that the licensing, build and distribution are significant issues. We are going to close this PR and make the code available in a separate repo where we can make the appropriate changes quickly without too much disturbance to this repo.

@matrober-uk we will ping you on the new location asap

PR #503, opened 2019-08-02
Updating pkg to latest for 0.8 release.

n3wscott:
## Proposed Changes

- Prepping for 0.8 release. Update knative.dev/pkg to latest from branch release-0.8.

**Release Note**

```release-note
NONE
```

/cc @mattmoor


Gopkg.toml
```diff
@@ -37,7 +37,7 @@ required = [
 
 [[override]]
   name = "knative.dev/pkg"
-  branch = "master"
+  branch = "release-0.8"
```
grantr: Just so I understand, do we expect to switch this back after the 0.8 cut next week?

Gopkg.toml
```diff
@@ -104,3 +104,22 @@ required = [
 [[constraint]]
   name = "github.com/apache/camel-k"
   version = "0.3.4"
+  
+
+# The dependencies below are required for opencensus.
```
grantr: Can this be "The next 4 dependencies" so if someone appends a new one it won't get added to the list? Or copy this comment to each override just to be safe.

PR #507, opened 2019-08-05
Automatic release tag labels

grantr:
Companion to https://github.com/knative/eventing/pull/1644.

## Proposed Changes

- Automatically update version labels in release artifacts when a tagged release is generated
- Add release labels to non-ConfigMap objects that don't have them, mainly Kafka and Camel sources.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -32,11 +32,20 @@ COMPONENTS=(
 readonly COMPONENTS
 
 function build_release() {
+   # Update release labels if this is a tagged release
+  if [[ -n "${TAG}" ]]; then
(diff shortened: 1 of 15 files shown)


hack/release.sh
```diff
@@ -32,11 +32,20 @@ COMPONENTS=(
 readonly COMPONENTS
 
 function build_release() {
+   # Update release labels if this is a tagged release
+  if [[ -n "${TAG}" ]]; then
```
nachocano: who sets this TAG env variable?

grantr: It's set by the release helpers when both `--tag-release` and `--version` are set.

PR #516, opened 2019-08-07
release patch: renaming kafka to kafka importer, and adding CRD and CCP based channels

### LINKED ISSUE #512: Kafka Channel was not released in 0.8.0
## Expected Behavior

KafkaChannel v0.8.0 can be installed from a [release artifact](https://github.com/knative/eventing-contrib/releases/tag/v0.8.0).

## Actual Behavior

KafkaChannel was not released for v0.8.0 because [release.sh only knows about the KafkaSource](https://github.com/knative/eventing-contrib/blob/4bda4fcaa57d4a344ea5a3a66e8c1a2ea2f04fda/hack/release.sh).

Looks like https://github.com/knative/eventing-contrib/pull/499 missed updating hack/release.sh.

/cc @matzew

matzew:
Fixes #512

## Proposed Changes

  * adding kafka channels (CRD / CCP) to release script (#512)
  *
  *

**Release Note**

```release-note
 include the missing kafka channels
```

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -26,7 +26,9 @@ COMPONENTS=(
   ["gcppubsub.yaml"]="gcppubsub/config"
   ["event-display.yaml"]="config/tools/event-display"
   ["camel.yaml"]="camel/source/config"
-  ["kafka.yaml"]="kafka/source/config"
+  ["kafka-importer.yaml"]="kafka/source/config"
+  ["kafka-channel-crd.yaml"]="kafka/channel/config"
+  ["kafka-channel-ccp.yaml"]="kafka/channel/config/provisioner"
   ["awssqs.yaml"]="contrib/awssqs/config"
 )
 readonly COMPONENTS


matzew:
@grantr and @vaikas-google here is the patch for the 0.8 release branch

hack/release.sh
```diff
@@ -26,7 +26,9 @@ COMPONENTS=(
   ["gcppubsub.yaml"]="gcppubsub/config"
   ["event-display.yaml"]="config/tools/event-display"
   ["camel.yaml"]="camel/source/config"
-  ["kafka.yaml"]="kafka/source/config"
+  ["kafka-importer.yaml"]="kafka/source/config"
+  ["kafka-channel-crd.yaml"]="kafka/channel/config"
```
grantr: Can we just call this `kafka-channel.yaml` so it doesn't change names when the CCP goes away?

grantr: See https://github.com/knative/eventing-contrib/pull/515/files#diff-cede583a27c227d458ca67079957c8a2R30

matzew: on the master we have it named `kafka-channel`, I felt its better to make it clear here, but happy to rename it

PR #491, opened 2019-07-27
Add Nats channel CRD implementation

montaro:
Fixes https://github.com/knative/eventing/issues/1567

**Release Note**
```release-note
Added Nats cluster channel provisioner implementation
```

PR #517, opened 2019-08-07
Adding NatssChannel CRD for release script

matzew:
## Proposed Changes

  * adding natss channel (CRD) to release script (was not added with #491)

**Release Note**

```release-note
```

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -28,6 +28,7 @@ COMPONENTS=(
   ["camel.yaml"]="camel/source/config"
   ["kafka-importer.yaml"]="kafka/source/config"
   ["kafka-channel.yaml"]="kafka/channel/config"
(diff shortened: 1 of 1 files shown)


hack/release.sh
```diff
@@ -28,6 +28,7 @@ COMPONENTS=(
   ["camel.yaml"]="camel/source/config"
   ["kafka-importer.yaml"]="kafka/source/config"
   ["kafka-channel.yaml"]="kafka/channel/config"
+  ["natss.yaml"]="natss/config"
```
grantr: Should this be called `natss-channel.yaml`? I know we don't have a NATSS importer, but it might be less confusing to users if all channels have the `-channel` suffix.

grantr: /hold

matzew:
/cancel hold

@grantr updated! good point!

PR #539, opened 2019-08-12
Don't ignore parsing errors

duglin:
If the request generates an error (e.g. HMAC verification failed) we shouldn't
just silently ignore it, the event will be nil so there's no point in going on.
We should return/log the error so someone can figure out what's going on.

Signed-off-by: Doug Davis <dug@us.ibm.com>

/cc @n3wscott


lionelvillard:
@duglin can you rebase?

/lgtm

duglin:
done

PR #540, opened 2019-08-12
Add kafka-source to addressable resolver imc

lberk:
While trying to wire up a `KafkaSource` to a `InMemoryChannel` was getting errors from the kafka-controller-manager.

`kubectl auth can-i get inmemorychannel.messaging.knative.dev --as=system:serviceaccount:knative-sources:kafka-controller-manager`
would return `no`

Adding this clusterrole fixed things.


Harwayne:
/hold

If this is required, then it should be in knative/eventing, where InMemoryChannel is installed, rather than here.

lberk:
> /hold
> 
> If this is required, then it should be in knative/eventing, where InMemoryChannel is installed, rather than here.

sure thing, will investigate a bit more as to why that didn't seem to work in this case

Issue #562, opened 2019-08-29
GitHubSource shouldn't use Service v1beta1 yet


grantr:
Doesn't work reliably yet on <1.13.10 k8s.

grantr:
See also https://github.com/knative/eventing-contrib/pull/489 and https://github.com/knative/eventing-contrib/issues/481

PR #489, opened 2019-07-23
Added v1beta1 serving API for github source.

chhsia0:
Fixes #481 

## Proposed Changes

  * Added the v1beta1 serving API to the scheme instead of v1alpha1.

**Release Note**

```NONE
```

PR #565, opened 2019-08-30
Revert upgrade to Beta serving. Use Alpha api but Beta shape. Lemonaid.

n3wscott:
Fixes https://github.com/knative/eventing-contrib/issues/562

## Proposed Changes

  * Use Alpha api but Beta shape. Lemonaid.

**Release Note**

```release-note
Github will launch as 1valpha1 serving resource.
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1811,6 +1811,7 @@
     "knative.dev/pkg/test",
     "knative.dev/pkg/test/logging",
     "knative.dev/pkg/webhook",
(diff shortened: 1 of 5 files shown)


grantr:
What about https://github.com/knative/eventing-contrib/blob/master/github/pkg/apis/addtoscheme_serving_v1beta1.go?

Does this change trigger https://github.com/knative/eventing-contrib/issues/481, which was the original reason for https://github.com/knative/eventing-contrib/pull/489 which updated to v1beta1?

PR #576, opened 2019-09-06
eventing-contrib test work with go1.13

chaodaiG:
Test would fail with go1.13 as it changed where testing flags are registered, fixing it by deleting explicit flag parsing like in https://github.com/knative/serving/pull/5398

/cc @mattmoor 
/cc @n3wscott 
/cc @adrcunha

### DIFF
--- a/test/e2e_flags.go
+++ b/test/e2e_flags.go
@@ -19,8 +19,6 @@ limitations under the License.
 package test
 
 import (
-	"flag"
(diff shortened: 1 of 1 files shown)


adrcunha:
```
# knative.dev/eventing-contrib/test
test/e2e_flags.go:22:2: imported and not used: "flag"
```

PR #591, opened 2019-09-13
Fix pkg

sayanh:
## Proposed Changes

  * Fix pkg


Fredy-Z:
Seems it does nothing except adding OWNERS files? And these files should be cleaned up by running `./hack/update-deps.sh`

PR #593, opened 2019-09-16
Documentation Improvements

nachtmaar:
## Proposed Changes

  * Adjust all commands to use flat-hierarchy (introduced [here](https://github.com/knative/eventing-contrib/blob/master/DEVELOPMENT.md#installing-sources)) instead of `contrib/<source_name>` since `contrib` folder does not exist anymore
  * Fix kubectl command using source=heartbeats-sender label
  * Migrate channels and subscriptions to messaging.knative.dev
  * Fix markdown linter and dead links complaints

**Release Note**

```release-note
Improve documentation of eventing-contrib
```

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -89,7 +89,7 @@ kubectl -n knative-sources logs \
 ```
 
 _See [camel/source/samples/README.md](./camel/source/samples/README.md),
-[kafka/source/samples/README.md](./kafka/source/samples/README.md) for
+[kafka/source/README.md](./kafka/source/README.md) for
 instructions on installing the Camel Source and Kafka Source._
 
 ## Iterating
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -79,7 +79,13 @@ github-controller-manager-0   1/1       Running   0          2h
 You can access the Github eventing manager's logs with:
 
 ```shell
-kubectl -n knative-sources logs $(kubectl -n knative-sources get pods -l control-plane=github-controller-manager -o name)
+kubectl -n knative-sources logs \
+    $(kubectl \
+        -n knative-sources \
+        get pods \
+        -l control-plane=github-controller-manager \
+        -o name \
+    )
 ```
 
 _See [camel/source/samples/README.md](./camel/source/samples/README.md),
@@ -92,7 +98,7 @@ As you make changes to the code-base:
 
 - **If you change a package's deps** (including adding external dep), then you
   must run [`./hack/update-deps.sh`](./hack/update-deps.sh).
-- **If you change a type definition (<source_name>/pkg/apis/),** then
+- **If you change a type definition (\<source_name\>/pkg/apis/),** then
   you must run [`./hack/update-codegen.sh`](./hack/update-codegen.sh). _This
   also runs [`./hack/update-deps.sh`](./hack/update-deps.sh)._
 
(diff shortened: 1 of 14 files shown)


DEVELOPMENT.md
```diff
@@ -62,7 +62,7 @@ of the sources _Github Source_, _AWS SQS Source_, _Camel Source_, _Kafka Source_
 with:
 
 ```
-ko apply -f contrib/<source_name>/
+ko apply -f <source_name>/config  # e.g. github/config
```
matzew: above in the document, there is an outdated link to the `ko` page on github.

Can you update that, so it than points to https://github.com/google/ko   ?

matzew: can you also add a statement to do the proper KO export,

like:  `export KO_DOCKER_REPO="docker.io/USERNAME"`

awssqs/samples/README.md
```diff
@@ -11,21 +11,21 @@ commands from the root.
...
+1. Setup
+   [Knative Eventing](https://github.com/knative/docs/tree/master/eventing).
 
-1.  The
-    [in-memory `ClusterChannelProvisioner`](https://knative.dev/eventing/tree/master/config/provisioners/in-memory-channel)
-    should be installed in your cluster. At the time of writing (release v0.5.0)
-    it is part of the default instructions so you will probably have it there
-    already.
+1. The
+   [in-memory `ClusterChannelProvisioner`](https://knative.dev/eventing/tree/master/config/provisioners/in-memory-channel)
```
matzew: that thing is gone. CCP is removed, and is replaced by the CRD based channels

matzew: Please update the yaml example files too.  Eg. w/ the new 0.9.0 release (coming on 18th Sep.), the `Subscription` is in the `messaging.knative.dev/v1alpha1` group

Also the ksvc uses older syntax, worth to update

nachtmaar: True, I see more occurences of `CCP` in the `eventing-contrib` repo. Are you ok with a separate PR for this? I already removed CCP from `docs` and `eventing` repo. Seems I forgot this one.

DEVELOPMENT.md
nachtmaar: Sure. Done

awssqs/samples/README.md
nachtmaar: Can do it immediately.

nachtmaar: @matzew

nachtmaar: We decided to change the context of the PR a bit and do the changes in one PR.

awssqs/samples/display-resources.yaml
```diff
@@ -1,26 +1,25 @@
 # Channel for testing events.
 
-apiVersion: eventing.knative.dev/v1alpha1
+apiVersion: messaging.knative.dev/v1alpha1
 kind: Channel
```
matzew: why not directly using the `InMemeoryChannel` kind here ?

nachtmaar: For no real reason. Is there a benefit of using it directly? Or is there a knative preference of using the concrete CRD directly instead of the abstract channel? We can change it if you want, but then let's do it for all channels in `eventing-contrib` for the sake of consistency. What do you think?

matzew: generic channel is OK - I was just wondering, why

PR #601, opened 2019-09-17
Improvements to KafkaSource adapter

slinkydeveloper:
## Proposed Changes

  * Added memory pool to reuse memory for created events
  * Removed useless marshalling/unmarshalling of json

### DIFF
--- a/kafka/source/pkg/adapter/adapter.go
+++ b/kafka/source/pkg/adapter/adapter.go
@@ -77,7 +77,14 @@ func (a *Adapter) Handle(ctx context.Context, msg *sarama.ConsumerMessage) (bool
 		return true, nil // Message is malformed, commit the offset so it won't be reprocessed
 	}
 
-	event := a.eventsPool.Get().(*cloudevents.Event)
+	var event *cloudevents.Event
+	if poolRes := a.eventsPool.Get(); poolRes != nil {
+		event = poolRes.(*cloudevents.Event)
+	} else {
+		event = &cloudevents.Event{}
+		event.SetSpecVersion(cloudevents.CloudEventsVersionV02)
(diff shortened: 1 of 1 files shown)


kafka/source/pkg/adapter/adapter.go
```diff
@@ -134,6 +139,17 @@ func (a *Adapter) Start(ctx context.Context, stopCh <-chan struct{}) error {
 		kafkaConfig.Net.TLS.Config = tlsConfig
 	}
 
+	// Create the events pool
+	a.eventsPool = sync.Pool{
+		New: func() interface{} {
+			// The Pool's New function should generally only return pointer
+			// types, since a pointer can be put into the return interface
+			// value without an allocation:
+			val := cloudevents.New(cloudevents.CloudEventsVersionV02)
```
matzew: this is likely something to be done for the channel too  ? 

(since both use the same common/handler for consumption)

slinkydeveloper: In channel probably the optimization could be different, i need to check it. Anyway, i think it's out of the scope of this pr

slinkydeveloper:
Before:

```
BenchmarkHandle/Baseline-4              10000000               255 ns/op             163 B/op          2 allocs/op
BenchmarkHandle/Benchmark_Handle-4         10000            235736 ns/op           19476 B/op        297 allocs/op
```

Now:

```
BenchmarkHandle/Baseline-4              10000000               265 ns/op             163 B/op          2 allocs/op
BenchmarkHandle/Benchmark_Handle-4         10000            205500 ns/op           16158 B/op        258 allocs/op
```

Benchmark pr incoming

PR #631, opened 2019-09-26
Source registry

nachocano:
Helps with https://github.com/knative/eventing/issues/1550

## Proposed Changes

  * Moving registry information to an annotation

**Release Note**

```release-note
NONE
```


nachocano:
how come this affects the natss e2e?

/test pull-knative-eventing-contrib-integration-tests

nachocano:
the problem is that the e2e use eventing head, and there has been some changes to the channel dispatcher and receiver... will need to fix that in order for this to pass...

PR #650, opened 2019-10-15
update eventing dep

lionelvillard:
## Proposed Changes

  * update dependencies


couchdb/source/pkg/reconciler/controller.go
```diff
@@ -60,7 +59,6 @@ func NewController(
 	}
 	impl := controller.NewImpl(r, r.Logger, ReconcilerName)
 
-	r.resourceTracker = resourceInformer.NewTracker(impl.EnqueueKey, controller.GetTrackerLease(ctx))
```
nachocano: thanks for taking care of this @lionelvillard!

PR #651, opened 2019-10-15
couchdb uses shared adapter main

lionelvillard:
## Proposed Changes

  * consolidate couchdb adapter main code.

### DIFF
--- a/couchdb/source/pkg/adapter/adapter.go
+++ b/couchdb/source/pkg/adapter/adapter.go
@@ -18,68 +18,87 @@ package adapter
 
 import (
 	"context"
+	"io/ioutil"
 	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"knative.dev/eventing/pkg/adapter"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/source"
 
 	"knative.dev/eventing-contrib/couchdb/source/pkg/apis/sources/v1alpha1"
 	"knative.dev/eventing-contrib/couchdb/source/pkg/couchdb"
 )
 
-// Hold options for the adapter
-type Options struct {
-	EventSource string
-	CouchDbURL  string
-	Username    string
-	Password    string
-	Database    string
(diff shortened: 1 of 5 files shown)


couchdb/source/pkg/adapter/adapter.go
```diff
@@ -18,68 +18,87 @@ package adapter
...
 		changeArgs: couchdb.ChangeArguments{},
 	}
-
-	return a, nil
 }
 
-func (a *adapter) Start(stopCh <-chan struct{}) error {
+func (a *couchDbAdapter) Start(stopCh <-chan struct{}) error {
 	// Just Poll for now. Other modes (long running, continues) will be provided later on.
 	return wait.PollUntil(2*time.Second, a.send, stopCh)
```
matzew: why is that `PollUntil` ?

couchdb/source/pkg/adapter/adapter.go
```diff
@@ -18,68 +18,87 @@ package adapter
...
+	rawusername, err := ioutil.ReadFile(env.CouchDbCredentialsPath + "/username")
+	if err != nil {
+		logger.Fatal("Missing username key in secret", zap.Error(err))
 	}
 
-	a := &adapter{
-		ce:         ceClient,
-		logger:     logger,
-		namespace:  options.Namespace,
+	rawpassword, err := ioutil.ReadFile(env.CouchDbCredentialsPath + "/password")
```
matzew: Not for now, but open issue to support secrets ?

matzew: Kafka has that - and we could see to share some ?

couchdb/source/pkg/adapter/adapter.go
matzew: Our PromQL does similar - I wonder if that generally can be improved   ( not saying lije CronJob .. .but ... :-) )

couchdb/source/pkg/adapter/adapter.go
lionelvillard: the adapter deployment mounts the secret.

lionelvillard:
@vaikas-google can you approve?

PR #684, opened 2019-10-29
CouchDB sink now uses Destination

lionelvillard:
## Proposed Changes

  * CouchDB sink is now a Destination


matzew:
/hold

wait until this also has merged for CORE sources.

I guess that means we do this in 0.11 ?

grantr:
@matzew this change has already merged for core sources right?

matzew:
/hold cancel 

core was already merged - sorry! 
just unfortunate timing, so close to the release .... bad coordination, IMO ...

matzew:
> feels like it was just quickly pushed through ... disappointed

not this one, more on the _eventing core_ ....

lionelvillard:
/hold cancel

Added CRD changes

PR #688, opened 2019-10-29
Kafka src desitation

matzew:
Fixes #

## Proposed Changes

  * KafkaSource sink now uses Destination
  *
  *

**Release Note**

```release-note
KafkaSource sink now uses Destination
```


kafka/source/pkg/reconciler/kafkasource.go
```diff
@@ -75,7 +76,8 @@ type Reconciler struct {
 	receiveAdapterImage string
 	eventTypeLister     eventinglisters.EventTypeLister
 	kafkaLister         listers.KafkaSourceLister
-	sinkReconciler      *duck.SinkReconciler
+	sinkResolver     	*resolver.URIResolver
+	sinkReconciler   	*duck.SinkReconciler
```
vaikas-google: I don't think this should be there anymore? sinkResolver should be the only thing you need?

kafka/source/config/300-kafkasource.yaml
```diff
@@ -126,7 +126,30 @@ spec:
...
+                          minLength: 1
+                        kind:
+                          type: string
+                          minLength: 1
+                        name:
+                          type: string
+                          minLength: 1
+                    uri:
+                      type: string
+                      description: "the target URI. If ref is provided, this must be relative URI reference."
```
grantr: You probably want https://github.com/knative/eventing/blob/86b99cb7741934bca31e6b862356f62c43e57c1b/config/300-apiserversource.yaml#L89-L104 also

matzew: ok, @grantr will PR the same for couch as well - after these are in, ok ?

grantr: :+1:

lionelvillard: I'm on it

grantr: @lionelvillard I think we should also change `anyOf` to `oneOf`

PR #701, opened 2019-10-31
Pull in the latest knative/pkg, incl. webhook refactor.

mattmoor:
This pulls in a series of changes to knative.dev/pkg/webhook that allows the boilerplate we need to be greatly reduced. The webhook is also now reconciler-based, so it should be much more resilient.


mattmoor:
cc @matzew 

This is now rebased, let's hope the OOMs are gone.

PR #729, opened 2019-11-14
Prometheus Source: crontab-formatted schedule and range queries

syedriko:
## Proposed Changes
- Added crontab-formatted schedule property to the Prometheus Source CR
- Added support for PromQL range queries

### DIFF
--- a/prometheus/pkg/adapter/adapter.go
+++ b/prometheus/pkg/adapter/adapter.go
@@ -22,12 +22,11 @@ import (
 	"crypto/x509"
 	"io/ioutil"
 	"net/http"
-	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go"
+	"github.com/robfig/cron"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/util/uuid"
-	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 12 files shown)


prometheus/pkg/adapter/adapter.go
```diff
@@ -83,56 +85,50 @@ func NewAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, ceClie
...
-		if !caCertPool.AppendCertsFromPEM(caCert) {
-			a.logger.Error("Error parsing CA certificate from " + caCertFile)
-			return err
-		}
-		a.client.Transport = &http.Transport{
-			TLSClientConfig: &tls.Config{
-				RootCAs: caCertPool,
-			},
-		}
+	sched, err := cron.ParseStandard(a.schedule)
```
matzew: nice to align w/ the cron package.

the `CronjobSource` does similar, but also reflects the schedule in the status. see:

https://github.com/knative/eventing/blob/master/pkg/reconciler/cronjobsource/cronjobsource.go#L168-L173

Should we do the same ?

syedriko: I think that's a fine idea. Please see the next commit.

PR #737, opened 2019-11-15
Prometheus source: added an example for Knative Monitoring

syedriko:
## Proposed Changes
Updated the README and added a scenario for how the Prometheus Source can be configured to communicate with the Prometheus server deployed as part of Knative Monitoring.

### DIFF
--- a/prometheus/README.md
+++ b/prometheus/README.md
@@ -3,81 +3,157 @@
 The Prometheus Event Source enables Knative Eventing integration with the
 [Prometheus Monitoring System](https://prometheus.io/).
 
-At the moment the implementation supports only simple PromQL queries. The
-queries run against the Prometheus server every 5 seconds and the results are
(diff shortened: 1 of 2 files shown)


prometheus/README.md
```diff
@@ -3,48 +3,18 @@
...
 [Prometheus Monitoring System](https://prometheus.io/).
 
-At the moment the implementation supports only simple PromQL queries. The
-queries run against the Prometheus server every 5 seconds and the results are
-sent to the configured sink as CloudEvents.
+## Instant and Range Queries
 
-## Deployment steps
-
-### Kubernetes
```
matzew: I was still liking that as an example to external source - let's keep it

syedriko: Added a section on off-cluster Prometheus server, please see the next commit.

PR #749, opened 2019-11-22
Follow-up: service ports name following istio-naming convention

iamejboy:
To be on safer side when using hand in hand with Istio, we follow service port naming convention as stated here ->https://istio.io/docs/setup/kubernetes/additional-setup/requirements/
Serving/Eventing already updated.

We have seen issue with istio as been discussed over here -> istio/istio#14520

Fixes #

## Proposed Changes

- Service port name should follow the istio naming convention based on this link:

https://istio.io/docs/setup/kubernetes/additional-setup/requirements/

**Release Note**

```
* Service port name should follow the istio naming convention based on this link:
https://istio.io/docs/setup/kubernetes/additional-setup/requirements/
```

### DIFF
--- a/awssqs/config/400-controller-service.yaml
+++ b/awssqs/config/400-controller-service.yaml
@@ -24,4 +24,5 @@ spec:
   selector:
     control-plane: awssqs-controller-manager
   ports:
-  - port: 443
+  - name: https-awssqs
+    port: 443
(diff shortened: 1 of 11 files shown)


iamejboy:
@googlebot I fixed it.

Issue #758, opened 2019-11-26
consumer group stops consuming regularly in Kafka channel dispatcher 


yuzisun:
## Expected Behavior
consumer group stops consuming after sometime 

## Actual Behavior
The consumer stops consuming in mid night every day and we had to restart the service, the Kafka channel status still reports fine

```
kubectl get kafkachannels -n xxx
NAME                         READY   REASON   HOSTNAME                                                        AGE
kf-ents-qc-input             True             kf-ents-qc-input-kn-channel.xxx.svc.cluster.local             27h

kubectl get subscriptions -n xxx
NAME                                       READY   REASON                                AGE
kf-ents-qc-input-es-lacus-out-to-parquet   True                                          27h
```
```
"level":"warn","ts":"2019-11-26T02:02:09.036Z","logger":"controller","caller":"dispatcher/dispatcher.go:213","msg":"Error in consumer group" 
"level":"info","ts":"2019-11-26T02:02:09.046Z","logger":"controller","caller":"kafka/consumer_handler.go:85","msg":"Stopping Consumer Group Handler, topic: kf-ents-qc-input" 
"level":"info","ts":"2019-11-26T02:02:09.145Z","logger":"controller","caller":"kafka/consumer_handler.go:85","msg":"Stopping Consumer Group Handler, topic: kf-ents-qc-input" 
"level":"info","ts":"2019-11-26T02:02:09.230Z","logger":"controller","caller":"kafka/consumer_handler.go:85","msg":"Stopping Consumer Group Handler, topic: kf-ents-qc-input" 
{"level":"info","ts":"2019-11-26T02:02:11.568Z","logger":"controller","caller":"kafka/consumer_handler.go:80","msg":"Message marked: , timestamp = 2019-11-26 02:01:05.715 +0000 UTC, topic = kf-ents-qc-input"}
"level":"info","ts":"2019-11-26T02:02:11.568Z","logger":"controller","caller":"kafka/consumer_handler.go:85","msg":"Stopping Consumer Group Handler, topic: kf-ents-qc-input"
{"level":"warn","ts":"2019-11-26T02:02:11.568Z","logger":"controller","caller":"dispatcher/dispatcher.go:213","msg":"Error in consumer group","error":"Unable to complete request unexpected HTTP response,
expected 2xx, got 502"}
[...]

PR #762, opened 2019-11-28
Use Namespace and name on kafka consumer group

matzew:
In the past we used to use namespace/name on the consumergroup:
* https://github.com/knative/eventing/blob/release-0.5/contrib/kafka/pkg/dispatcher/dispatcher.go#L171

But https://github.com/knative/eventing/pull/1057 changed that to only use a `UID` on it.

**Problem:**

(re)creating subscriptions is ending up in different consumer groups.

The issue was raised by @yuzisun chatting, while looking at #758

## Proposed Changes

  * use a more reliable consumer group setup

**Release Note**

```release-note
Consumer groups will no longer have random UID (subscription UID) names, they will go back to the old behavior of using a stable combination of namespace and name
```

### DIFF
--- a/kafka/channel/pkg/dispatcher/dispatcher.go
+++ b/kafka/channel/pkg/dispatcher/dispatcher.go
@@ -75,6 +75,8 @@ func (c consumerMessageHandler) Handle(ctx context.Context, message *sarama.Cons
 var _ kafka.KafkaConsumerHandler = (*consumerMessageHandler)(nil)
 
 type subscription struct {
+	Namespace     string
+	Name          string
 	UID           string
 	SubscriberURI string
 	ReplyURI      string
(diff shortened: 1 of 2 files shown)


slinkydeveloper:
So that means that everytime a new subscription subscribe to the channel, it won't restart processing from beginning right?

yuzisun:
> So that means that everytime a new subscription subscribe to the channel, it won't restart processing from beginning right?

yes, for a given subscription the consumer group name should be stable. If a subscription is deleted and recreated it should not create a new consumer group(currently it is using the subscription UID) and consume again from beginning.

slinkydeveloper:
cool /lgtm

matzew:
/hold

we should NOT use the channel name ... will fix this

yuzisun:
> /hold
> 
> we should NOT use the channel name ... will fix this

good catch ! should be the subscription name

matzew:
With this, we now get exactly stable consumer groups, like:

`kafka DOT namespace DOT channelName DOT subscriptionName`

PR #766, opened 2019-11-30
Use K_SINK for where to send events.

mattmoor:


mattmoor:
I also added a ONE_SHOT mode to have it send a single event and quit (I did this in the knative/eventing version to facilitate CronJob-based e2e testing).

PR #788, opened 2019-12-10
KafkaChannel Dispatcher Refactor to use CloudEvents

lberk:
As per knative/eventing-contrib/issues/597 - refactored the kafkachannel to use Events vs messages.


lberk:
Still getting tests sorted
/wip

kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -303,33 +320,80 @@ func (d *KafkaDispatcher) getChannelReferenceFromHost(host string) (eventingchan
...
+	addHeader(&message, "ce_id", event.ID())
+	addHeader(&message, "ce_time", event.Time().Format(time.RFC3339))
+	if event.DataContentType() != "" {
+		addHeader(&message, "ce_datacontenttype", event.DataContentType())
+	}
+	if event.Subject() != "" {
+		addHeader(&message, "ce_subject", event.Subject())
+	}
+	if event.DataSchema() != "" {
+		addHeader(&message, "ce_dataschema", event.DataSchema())
```
matzew: I think they want `content-type` here when mapping to kafka ?

kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -303,33 +320,80 @@ func (d *KafkaDispatcher) getChannelReferenceFromHost(host string) (eventingchan
...
+	addHeader(&message, "ce_specversion", event.SpecVersion())
+	addHeader(&message, "ce_type", event.Type())
+	addHeader(&message, "ce_source", event.Source())
+	addHeader(&message, "ce_id", event.ID())
+	addHeader(&message, "ce_time", event.Time().Format(time.RFC3339))
+	if event.DataContentType() != "" {
+		addHeader(&message, "ce_datacontenttype", event.DataContentType())
+	}
+	if event.Subject() != "" {
+		addHeader(&message, "ce_subject", event.Subject())
```
matzew: there is no subject mentioned here:
https://github.com/cloudevents/spec/blob/v1.0/kafka-protocol-binding.md#325-example

@lberk can u follow up w/ an issue on the CE spec, for kafka? - looks like their 1.0 lacks some attrs for Kafka, like `subject`

kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -270,23 +279,31 @@ func NewDispatcher(args *KafkaDispatcherArgs) (*KafkaDispatcher, error) {
 		return nil, fmt.Errorf("unable to create kafka producer: %v", err)
 	}
 
+	ceClient, err := kncloudevents.NewDefaultClient()
```
nachocano: not needed for this PR, but I think it would be worth to look at the good_client in eventing, and see if we can use that one, or update the one in eventing_contrib... https://github.com/knative/eventing/blob/master/pkg/kncloudevents/good_client.go
I added some tuning there, where you can better reuse connections, and that improved perf quite a bit... 
something to keep in mind...

kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -303,33 +320,80 @@ func (d *KafkaDispatcher) getChannelReferenceFromHost(host string) (eventingchan
...
+	}
+	if event.Subject() != "" {
+		addHeader(&message, "ce_subject", event.Subject())
+	}
+	if event.DataSchema() != "" {
+		addHeader(&message, "ce_dataschema", event.DataSchema())
+	}
+	// Only setting string extensions.
+	for k, v := range event.Extensions() {
+		if vs, ok := v.(string); ok {
```
nachocano: I think this changed in v1. IIRC all are strings now... take a look, maybe you don't need that type cast..

kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -303,33 +320,80 @@ func (d *KafkaDispatcher) getChannelReferenceFromHost(host string) (eventingchan
...
+		case "ce_time":
+			t, _ := time.Parse(time.RFC3339, v)
+			event.SetTime(t)
+		case "ce_subject":
+			event.SetSubject(v)
+		case "ce_dataschema":
+			event.SetDataSchema(v)
+		default:
+			// Extensions
+			event.SetExtension(strings.ToLower(h), strings.ToLower(v))
```
nachocano: This might break if the header is non-alphanumeric in the case of 1.0. 
I had to do some ugly check for our PubSub source here: https://github.com/google/knative-gcp/blob/master/pkg/pubsub/adapter/converters/pubsub.go#L54
you might need something similar..

nachocano:
Great work @lberk !!! Just few minor comments... I'll let @matzew approve once the tests pass, as he is our Kafka guy!

kafka/channel/pkg/dispatcher/dispatcher.go
lberk: https://github.com/cloudevents/sdk-go/blob/master/pkg/cloudevents/eventcontext_v1_reader.go#L81-L89
hm, looks to me like it still returns a `map[string]interface{}` perhaps it's still better to type cast for now

kafka/channel/pkg/dispatcher/dispatcher.go
lberk: https://github.com/knative/eventing-contrib/issues/796 created an issue to track this in it's own PR

PR #797, opened 2019-12-13
Update to latest pkg and eventing

n3wscott:
## Proposed Changes

  * Migrate nats and kafka to use api.URL internally.

**Release Note**

```release-note
NONE
```


n3wscott:
/assign @mattmoor 
This should fix the bots.

PR #803, opened 2019-12-17
Cherrypick appender

vaikas:
## Proposed Changes

  * Cherrypick #801 and #802 

**Release Note**

```release-note
Add appender as a release artifact, it's used by samples in knative.dev/docs so should be in public repo.
```


vaikas:
@googlebot I fixed it.

Issue #809, opened 2019-12-19
Clarify Kafka Consumer Groups needs.


vaikas:
As part of cleaning up the old Deprecated fields from Subscription in https://github.com/knative/eventing/pull/2335

And then figuring out if there are code that uses it, I found this:
https://github.com/knative/eventing-contrib/blob/master/kafka/channel/pkg/dispatcher/dispatcher.go#L118

And I'm having a hard time understanding if the Subscription Name/Namespace is what's important here and why. If it's that important, maybe we should not then have deprecated this from the API surface.
I do not believe there's a way to link from a UID to the actual subscription (well, in a sane way without spinning through all the subscriptions and filtering by UID), so if there's something that fundamentally needs to tie the subscription to the actual Subscription resource, we should understand it.
@matzew Since you added that fairly recently, could we figure out if there's something else that we need to surface in the type?
https://github.com/knative/eventing-contrib/commit/9639631fa788887c118692be66c7b09ae37a0a4b

matzew:
see https://github.com/knative/eventing-contrib/pull/762 

In Kafka, when you have `n` partitions of a topic, you can have up `n` consumers for that one topic, in a (named) consumer group.

In order to have stable consumers (belonging to a consumer group), for that subscription it is important to have meaningful/stable consumer group names.

We picked the namespace and the name of the subscription, as well as the name of the channel in order to guarantee that.

```
kafka.mynamespace.mychannel.sub1
kafka.mynamespace.mychannel.sub2
```

Consumer groups ensure than that there is no "reply" of the messages from before, using UID and than (re)creating subs, is kinda randomizing that, but always having different consumergroup names, for the same subscription.

So now, it is always `kafka.mynamespace.mychannel.sub1`, instead of `kafka.123` and than `kafka.234` when recreating the sub.

This was brought up by @yuzisun recently

vaikas:
Ok, I need to grok this little more, but it seems that there are now assumptions about naming subscriptions and especially about deleting/recreating the subscriptions that have knowledge about the implementation details of a particular channel. As in, is it correct to state that if you delete a subscription / recreate it, there's now some assumption about where the events "pick up" and start coming from?

vaikas:
[...]

PR #814, opened 2020-01-06
Kafka: can we just use the uid and not name

n3wscott:
## Proposed Changes

  * do not use name from `DeprecatedRef` of subscription, use uid for kafka subs.

`spec.DeprecatedRef` will be removed.

**Release Note**

```release-note
name changes to uid for subscriptions in kafka.
```

### DIFF
--- a/kafka/channel/pkg/dispatcher/dispatcher.go
+++ b/kafka/channel/pkg/dispatcher/dispatcher.go
@@ -117,7 +117,7 @@ func (d *KafkaDispatcher) UpdateKafkaConsumers(config *multichannelfanout.Config
 		for _, subSpec := range cc.FanoutConfig.Subscriptions {
 			// TODO: use better way to get the provided Name/Namespce for the Subscription
 			// we NEED this for better consumer groups
-			sub := newSubscription(subSpec, subSpec.DeprecatedRef.Name, subSpec.DeprecatedRef.Namespace)
+			sub := newSubscription(subSpec, string(subSpec.UID), cc.Namespace)
 			if _, ok := d.kafkaConsumerGroups[channelRef][sub]; !ok {
 				// only subscribe when not exists in channel-subscriptions map
 				// do not need to resubscribe every time channel fanout config is updated


grantr:
Why?

n3wscott:
the code uses a deprecated field that will be deleted in the next version.

matzew:
This is kinda conflicting with https://github.com/knative/eventing-contrib/pull/762

One reason for doing the above is that with UID, the name of the consumer group is not stable when (re)creating the subscription.

/cc @yuzisun

matzew:
Question is: Do we really need stable consumer group names? when we (re)create subscriptions? 

To some degree the usage of consumer group (for subscription) is an impl. detail. and it's an internal client only, that's being used to consume things from the topic for redelivery.

slinkydeveloper:
The question is: Do we have any specification of what the `Channel` should behave when a destination removes the subscription and subscribe again?

- With #762 it doesn't resend messages already sent
- With this one it resend everything from the beginning of the topic (which could not be the beginning of the events flow, since topics clears periodically old messages)

How the other implementations behaves?

n3wscott:
> How the other implementations behaves?

Others would treat a new subscription with the same name as a new and unique subscription. There is no recreating the original subscription. It got deleted and there for it is gone. This is only true for channels. In Pub/Sub, you can make a pull subscription directly and recreate with an existing subscription id and it would be the same, and it is independent of the name of that subscription. 


PubSub:

Channel makes a Topic and PullSubscriptions

But you can make Topic and PullSubscription directly and fill the required fields yourself. The channel impl makes the choice that new subscriptions are always unique.

slinkydeveloper:
@n3wscott Can we document that behaviour somewhere, like in the [`Channel` spec document](https://github.com/knative/eventing/blob/master/docs/spec/channel.md)?

n3wscott:
> @n3wscott Can we document that behaviour somewhere, like in the [`Channel` spec document](https://github.com/knative/eventing/blob/master/docs/spec/channel.md)?

I think it should be up to the channel task force... But this has not come up?

slinkydeveloper:
@n3wscott let's discuss about it at the next one :+1: I've added it in the document

matzew:
@n3wscott I've added this item to the agenda for the Thursday call on the channel Spec group !

yuzisun:
Kafka is a streaming logging system not a pubsub system, the message still stays on broker after it is consumed and each consumer group maintains its own offset so they can consume at different pace. I understand for a pubsub system we can use UID but for Kafka it is important to have a stable consumer group for a given destination. If UID is used it would resend all the messages over again from beginning offset when subscription is recreated which is a big no for production environment.

slinkydeveloper:
I agree with @yuzisun, IMO the behaviour should be: when the channel receives the message, all subscribers should receive __at least once__ the message; then the channel could eventually forget the message. In the particular case of Kafka, this translates into having a single consumer group __per channel__, where the offset is marked when the message is correctly delivered.

Again, I think this should be well defined in the `Channel` specification (including what we mean in this context for _message correctly delivered_.

grantr:
> a single consumer group **per channel**

Can you help me understand how a single consumer group per channel would work? Not saying it wouldn't, I'm just wondering, if I have many subscribers at different offsets, how does the channel implementation track the offsets for those subscribers?

matzew:
/lgtm

reference: was discussed in the channel WG call today. See note from @vaikas 
https://github.com/knative/eventing-contrib/issues/809#issuecomment-572643926

slinkydeveloper:
@grantr I was wrong because i didn't took into account [`auto.offset.reset = latest`](https://kafka.apache.org/documentation/#consumerconfigs) :smile: what I mean is: the behaviour I expect is that every time i subscribe, i don't restart processing from the beginning of the topic. But with the actual configuration of the consumer (the default one) the dispatcher starts processing from the latest message: https://github.com/Shopify/sarama/blob/master/config.go#L435

vaikas:
@slinkydeveloper is really there no way to specify that when a consumer is created to Kafka that it will start from 'now' not from the beginning?

yuzisun:
> @slinkydeveloper is really there no way to specify that when a consumer is created to Kafka that it will start from 'now' not from the beginning?

Let's say consumer group A consumed offset up to 10, now subscription is deleted and recreated it then creates a totally new consumer group B in which case current offset does not exist, yes we can specify I want to consume from the latest offset but at that time it might be already 15 and you would lose the messages between 10-15, what's why you need a stable consumer group which can resume from 10.

vaikas:
Right, so what I'm saying is that if we can rationalize for the sake of moving forward a following behaviour:
 1. When a new Subscription is created, it will get events from that point forward. So, in this case, you would start at 15.
 2. We have a way to specify Channel Implementation specific features in a more generic way. For the time being you could use annotations to _explicitly_ specify the consumer group you want (which is more flexible than implicitly (and with no way to override)).

Current assumption

yuzisun:
> Right, so what I'm saying is that if we can rationalize for the sake of moving forward a following behaviour:
> 
> 1. When a new Subscription is created, it will get events from that point forward. So, in this case, you would start at 15.
> 2. We have a way to specify Channel Implementation specific features in a more generic way. For the time being you could use annotations to _explicitly_ specify the consumer group you want (which is more flexible than implicitly (and with no way to override)).
> 
> Current assumption

+1,  I did not know you can specify consumer group via annotation :).

PR #835, opened 2020-01-17
Minior updates on 'correcting' JSON to camelCase...

matzew:
Related to
* https://github.com/knative/eventing/pull/2373
* https://github.com/knative/eventing/pull/2400

## Proposed Changes

  * going with cameCase JSON

### DIFF
--- a/github/pkg/apis/sources/v1alpha1/githubsource_types.go
+++ b/github/pkg/apis/sources/v1alpha1/githubsource_types.go
@@ -77,7 +77,7 @@ type GitHubSourceSpec struct {
 
 	// API URL if using github enterprise (default https://api.github.com)
 	// +optional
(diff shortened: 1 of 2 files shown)


vaikas:
I'd propose doing these in the v1beta1 as they are non-backwards compatible.

github/pkg/apis/sources/v1alpha1/githubsource_types.go
```diff
@@ -77,7 +77,7 @@ type GitHubSourceSpec struct {
 
 	// API URL if using github enterprise (default https://api.github.com)
 	// +optional
-	GitHubAPIURL string `json:"githubAPIURL,omitempty"`
```
n3wscott: These will be breaking API changes.

antoineco:
@matzew we can probably close this since we don't want to break APIs?

PR #839, opened 2020-01-21
Update the signature of ResolveChannelFromHostFunc to match it in the eventing repo

shefaliv:
Changes in the eventing repo: https://github.com/knative/eventing/pull/2389

Also updating the knative.dev/eventing dependency.

Resolves https://github.com/knative/eventing/issues/1952


shefaliv:
I'll squash the commits (and undo the change to gopkg.toml) once the eventing repo changes are in and I can just update dependencies from the master branch!

pkg/channel/message_receiver.go
```diff
@@ -211,10 +218,10 @@ func (r *MessageReceiver) fromHTTPHeaders(headers http.Header) map[string]string
 
 // ParseChannel converts the channel's hostname into a channel
 // reference.
-func ParseChannel(host string) (channel.ChannelReference, error) {
-	chunks := strings.Split(host, ".")
+func ParseChannel(host url.URL) (channel.ChannelReference, error) {
+	chunks := strings.Split(host.Path, ".")
 	if len(chunks) < 2 {
-		return channel.ChannelReference{}, fmt.Errorf("bad host format '%s'", host)
+		return channel.ChannelReference{}, fmt.Errorf("bad host format %v", host)
```
antoineco: Should be `%q` to preserve the original intention of quoting the output.

vendor/knative.dev/eventing/pkg/apis/sources/v1alpha1/ping_defaults.go
```diff
@@ -18,20 +18,12 @@ package v1alpha1
...
-			s.Spec.ChannelTemplate = channelTemplate
-		}
-	}
+func (s *PingSource) SetDefaults(ctx context.Context) {
 	s.Spec.SetDefaults(ctx)
 }
 
-func (ss *ParallelSpec) SetDefaults(ctx context.Context) {}
+func (ss *PingSourceSpec) SetDefaults(ctx context.Context) {
+	// TODO anything?
```
antoineco: How about clarifying with the author instead of merging this? We all know most TODOs stay in the code for ever 😉

natss/pkg/dispatcher/dispatcher.go
```diff
@@ -360,11 +361,11 @@ func (s *SubscriptionsSupervisor) ProcessChannels(ctx context.Context, chanList
...
 }
 
-func (s *SubscriptionsSupervisor) getChannelReferenceFromHost(host string) (eventingchannels.ChannelReference, error) {
+func (s *SubscriptionsSupervisor) getChannelReferenceFromHost(host url.URL) (eventingchannels.ChannelReference, error) {
 	chMap := s.getHostToChannelMap()
-	cr, ok := chMap[host]
+	cr, ok := chMap[host.Path]
 	if !ok {
-		return cr, fmt.Errorf("Invalid HostName:%q. HostName not found in any of the watched natss channels", host)
+		return cr, fmt.Errorf("Invalid HostName:%v. HostName not found in any of the watched natss channels", host)
```
antoineco: Why not preserving the quotes? It makes it easier to distinguish the input from the message IMO.

antoineco: (not related to your changes) I doubt the absence of space between `:` and the string was intentional. Shall we use the occasion to fix the format?

kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -318,11 +319,11 @@ func NewDispatcher(args *KafkaDispatcherArgs) (*KafkaDispatcher, error) {
...
 }
 
-func (d *KafkaDispatcher) getChannelReferenceFromHost(host string) (eventingchannels.ChannelReference, error) {
+func (d *KafkaDispatcher) getChannelReferenceFromHost(host url.URL) (eventingchannels.ChannelReference, error) {
 	chMap := d.getHostToChannelMap()
-	cr, ok := chMap[host]
+	cr, ok := chMap[host.Path]
 	if !ok {
-		return cr, fmt.Errorf("invalid Hostname:%s. Hostname not found in ConfigMap for any Channel", host)
+		return cr, fmt.Errorf("invalid Hostname:%v. Hostname not found in ConfigMap for any Channel", host)
```
antoineco: (not related to your changes) I doubt the absence of space between `:` and the string was intentional. Shall we use the occasion to fix the format?

pkg/channel/message_receiver.go
matzew: +1

natss/pkg/dispatcher/dispatcher.go
matzew: +1

matzew: +1  let's get this changed!

kafka/channel/pkg/dispatcher/dispatcher.go
matzew: +1

pkg/channel/message_receiver.go
shefaliv: The type of host has changed, so I get this error with %q: Errorf format %q has arg host of wrong type net/url.URL

I can manually add in quotes though!

natss/pkg/dispatcher/dispatcher.go
shefaliv: Done.

kafka/channel/pkg/dispatcher/dispatcher.go
shefaliv: Done.

kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -318,11 +319,11 @@ func NewDispatcher(args *KafkaDispatcherArgs) (*KafkaDispatcher, error) {
...
 }
 
-func (d *KafkaDispatcher) getChannelReferenceFromHost(host string) (eventingchannels.ChannelReference, error) {
+func (d *KafkaDispatcher) getChannelReferenceFromHost(host url.URL) (eventingchannels.ChannelReference, error) {
 	chMap := d.getHostToChannelMap()
-	cr, ok := chMap[host]
+	cr, ok := chMap[host.Path]
 	if !ok {
-		return cr, fmt.Errorf("invalid Hostname:%s. Hostname not found in ConfigMap for any Channel", host)
+		return cr, fmt.Errorf("invalid Hostname: '%v'. Hostname not found in ConfigMap for any Channel", host)
```
n3wscott: host is not longer just a host, it is a full uri so this comment and the variable name is out of date.

natss/pkg/dispatcher/dispatcher.go
```diff
@@ -360,11 +361,11 @@ func (s *SubscriptionsSupervisor) ProcessChannels(ctx context.Context, chanList
...
 }
 
-func (s *SubscriptionsSupervisor) getChannelReferenceFromHost(host string) (eventingchannels.ChannelReference, error) {
+func (s *SubscriptionsSupervisor) getChannelReferenceFromHost(host url.URL) (eventingchannels.ChannelReference, error) {
 	chMap := s.getHostToChannelMap()
-	cr, ok := chMap[host]
+	cr, ok := chMap[host.Path]
 	if !ok {
-		return cr, fmt.Errorf("Invalid HostName:%q. HostName not found in any of the watched natss channels", host)
+		return cr, fmt.Errorf("'Invalid HostName: '%v'. HostName not found in any of the watched natss channels", host)
```
n3wscott: same here, hostname is now a uri

pkg/channel/message_receiver.go
```diff
@@ -211,10 +218,10 @@ func (r *MessageReceiver) fromHTTPHeaders(headers http.Header) map[string]string
 
 // ParseChannel converts the channel's hostname into a channel
 // reference.
-func ParseChannel(host string) (channel.ChannelReference, error) {
-	chunks := strings.Split(host, ".")
+func ParseChannel(host url.URL) (channel.ChannelReference, error) {
+	chunks := strings.Split(host.Path, ".")
```
n3wscott: this is pulling the path, not the host. is that an issue?

shefaliv: It matches the eventing repo: https://github.com/knative/eventing/blob/702be2256abf1be34390b7eb4ea63f600d46ba8d/pkg/channel/event_receiver.go#L185

kafka/channel/pkg/dispatcher/dispatcher.go
shefaliv: What's the problem with the comment here? Do you prefer "invalid host" instead of "invalid hohstname"?

n3wscott: host == a uri, not a hostname.

shefaliv: URI is a type. "Host" and "Hostname" are just what we choose to call it (or not), so I don't quite understand your last comment. Can you please be more direct in what you'd like adjusted here? Do you prefer "invalid host" instead of "invalid hostname"?

shefaliv: I just adjusted it to include host.Path instead of the full host.

pkg/channel/message_receiver.go
antoineco: %q invokes String(), so you can use it safely.

kafka/channel/pkg/dispatcher/dispatcher.go
antoineco: How about just renaming the variable?

shefaliv: Sure, can do! I also adjusted it to be host.Path instead of the full host since this error message is right after an assertion on host.Path.

pkg/channel/message_receiver.go
grantr: @shefaliv and I looked into this because it was odd to me too. It works because of a quirk of [`url.Parse`](https://golang.org/pkg/net/url/#Parse). If a URI has no scheme, url.Parse assumes that's a relative uri and puts the string in Path. So this doesn't actually change any behavior even though it looks wrong. FWIW a unit test in eventing failed when we tested switching it to use Host, so there is coverage for this.

What we really want here is url.ParseRequestURI which assumes a scheme-less uri is absolute and puts the string in Host. But we'll change that in a followup to eventing and eventing-contrib.

PR #840, opened 2020-01-21
Update the dependency on sdk-go to match eventing repo

shefaliv:
PR in eventing repo: https://github.com/knative/eventing/pull/2385

Also updating eventing dependency as necessary for this update.

### DIFF
--- a/kafka/channel/pkg/dispatcher/dispatcher.go
+++ b/kafka/channel/pkg/dispatcher/dispatcher.go
@@ -353,7 +353,7 @@ func fromKafkaMessage(ctx context.Context, kafkaMessage *sarama.ConsumerMessage)
 			event.SetDataSchema(v)
 		default:
 			// Extensions
-			if IsAlphaNumericLowercaseLetters(h) {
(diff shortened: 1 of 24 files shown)


kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -353,7 +353,7 @@ func fromKafkaMessage(ctx context.Context, kafkaMessage *sarama.ConsumerMessage)
 			event.SetDataSchema(v)
 		default:
 			// Extensions
-			if IsAlphaNumericLowercaseLetters(h) {
+			if IsAlphaNumeric(h) {
```
grantr: This is a behavior change: previously a header with uppercase letters would not be set as an extension. Now it will (and I believe sdk-go will then lowercase it). @n3wscott @matzew any thoughts on this change?

shefaliv: The repo doesn't build without this change. It's due to this breaking change: https://github.com/cloudevents/sdk-go/pull/274

PR #873, opened 2020-01-30
Allow bootstrapServers to be configured by secret

lberk:
If spec.bootstrapServers isn't specified (made optional), take the value from the secret passed.

We had a request to make the kafkasource's bootstrap server configurable via secret so that a tool like a [Service Binding Operator](https://github.com/redhat-developer/service-binding-operator) could set the default bootstrap server cluster wide (but the user can still override it by setting their own bootstrapServer).  Would this design be appropriate or is there a preferable implementation to make a source's field more configurable?

## Proposed Changes

  * Allow the bootstrap server env var in the receive adapter to be set via secret
  * If the user specifies a `spec.bootstrapServer` it is taken as priority
  * Make `spec.bootstrapServers` optional


kafka/source/pkg/apis/sources/v1alpha1/kafka_types.go
```diff
@@ -100,9 +100,13 @@ type KafkaResourceSpec struct {
 // KafkaSourceSpec defines the desired state of the KafkaSource.
 type KafkaSourceSpec struct {
 	// Bootstrap servers are the Kafka servers the consumer will connect to.
-	// +required
+	// +optional
```
matzew: hrm...  I think we want to require this... b/c w/o this thing is not working

matzew: ah.  XOR :-)   `BootstrapServers` vs `BootstrapServersSecret`

got it!

kafka/source/pkg/reconciler/kafkasource.go
```diff
@@ -244,6 +244,10 @@ func (r *Reconciler) createReceiveAdapter(ctx context.Context, src *v1alpha1.Kaf
 		logging.FromContext(ctx).Error("error while converting metrics config to JSON", zap.Any("receiveAdapter", err))
 	}
 
+	if src.Spec.BootstrapServers == "" && src.Spec.BootstrapServersSecret.SecretKeyRef == nil {
+		logging.FromContext(ctx).Error("both BootstrapServers and BootstrapServersSecret are empty, one must be specified")
```
matzew: :+1:

matzew:
/assign @lionelvillard 
/assign @aslom 

Lionel/Alek. you guys talked about "more common" config/setup of Kafka - please take a look here, thanks!

lionelvillard:
/hold

@matzew is going to start drafting a proposal for Kafka v1beta1 and that might impact this PR.

matzew:
> @matzew is going to start drafting a proposal for Kafka v1beta1 and that might impact this PR.

@lionelvillard I generally like to see this being done before `v1beta1`, or do you mean that we might try to aim and unify the config for channel and source?

lionelvillard:
let's start with the CRD API and let's see how far we manage to go :-)

matzew:
One thing to think about is... do all props need secrets?  or are some just fine w/ ConfigMaps ?

lionelvillard:
most of the props should be in ConfigMap. Only the secrets should be in Secrets :-)

Issue #863, opened 2020-01-29
[KafkaSource] Migrate to sources.knative.dev api group


n3wscott:
## Expected Behavior
The resource is inside of the api group `sources.knative.dev`

## Actual Behavior
`sources.eventing.knative.dev`

/assign bbrowning

bbrowning:
This was solved by #875 and can be closed, right?

lberk:
Specifically #902 after the correct import
/close

PR #875, opened 2020-01-30
Source Migration for kafkaSource, promSource and githubSource

### LINKED ISSUE #864: [PrometheusSource] Migrate to sources.knative.dev api group
## Expected Behavior
The resource is inside of the api group `sources.knative.dev`

## Actual Behavior
`sources.eventing.knative.dev`

/assign syedriko

lberk:
Fixes #864 
#863
#862

### DIFF
--- a/github/config/201-clusterrole.yaml
+++ b/github/config/201-clusterrole.yaml
@@ -21,7 +21,7 @@ metadata:
 rules:
 # Sources admin
 - apiGroups:
-  - sources.eventing.knative.dev
+  - sources.knative.dev
   resources:
   - githubsources
   verbs: &everything
@@ -35,14 +35,14 @@ rules:
 
 # Sources finalizer
 - apiGroups:
-  - sources.eventing.knative.dev
+  - sources.knative.dev
   resources:
   - githubsources/finalizers
   verbs: *everything
 
 # Source statuses update
 - apiGroups:
-  - sources.eventing.knative.dev
+  - sources.knative.dev
   resources:
   - githubsources/status
   verbs:
@@ -102,7 +102,7 @@ metadata:
     duck.knative.dev/source: "true"
 rules:
   - apiGroups:
-      - "sources.eventing.knative.dev"
+      - "sources.knative.dev"
     resources:
       - "githubsources"
     verbs:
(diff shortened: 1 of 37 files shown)


matzew:
@lberk hrm... something wrong :-)

lionelvillard:
/hold

this change is not backward compatible.

lberk:
@lionelvillard While this isn't a backward compatible change, how would you suggest we approach the change?  double up the apigroups to have a `sources.eventing.knative.dev` in parallel to `eventing.knative.dev` for a release? (before dropping the former?)  I think @n3wscott changed eventing by moving apigroups to legacysources first (which in itself is a breaking change as well?)
@n3wscott @matzew thoughts?

lberk:
As an aside, isn't this part of the rational for still being in `v1alpha1` ? that we can still make changes like this to the apigroup names?

matzew:
> As an aside, isn't this part of the rational for still being in v1alpha1 ? that we can still make changes like this to the apigroup names?

yeh, I think so too

n3wscott:
> I think @n3wscott changed eventing by moving apigroups to legacysources first (which in itself is a breaking change as well?)

It is code breaking, API non-breaking. So old releases still work with the API in the ApiServer.

n3wscott:
My vote would be to allow in an API breaking change like this but I wanted to leave that choice to the owners of the sources. It is a ton of work to do what I had to do in eventing with the legacy client.

matzew:
> My vote would be to allow in an API breaking change like this but I wanted to leave that choice to the owners of the sources. It is a ton of work to do what I had to do in eventing with the legacy client.

Honestly, I feel exactly the same, and I see why you did the extra mile for eventing CORE.   

for GH/Prom/KafkaSource, I'd be OK to let it break.   At some point users have to migrate anyways.

Just my $0.02

Abd4llA:
I think at v1alpha1 unless we explicitly hear from owners or users of specific sources that they wanna avoid a breaking change, we should go ahead and do specially considering the overhead would be non beneficial.

lionelvillard:
I'm also fine with breaking changes.

lberk:
Thanks folks, if everybody's ok with the changes as is -- could I get a `/lgtm` ?

PR #878, opened 2020-01-31
Use ce overrides if given in heartbeat sample.

n3wscott:
## Proposed Changes

  * Use ce overrides in heartbeat sample.

**Release Note**

```release-note
Heartbeats sample uses CloudEvents Overrides if provided. See runtime contract for SinkBinding. for more details.
```


n3wscott:
Relayed to https://github.com/knative/eventing/pull/2481

PR #880, opened 2020-02-03
Fixing missing error handling in kafka consumer group handler

### LINKED ISSUE #851: kafka consumer_factory consumerGroup.Consume connection errors are ignored
https://github.com/knative/eventing-contrib/blob/86bac88600f79123e3e29babc33a44007c6ff419/kafka/common/pkg/kafka/consumer_factory.go#L80-L81

consumerGroup.Consume errors are ignored.

While the comment "// If any error, they are reported in error channel, so we can skip this message" suggests it is not needed, it is not exactly true, as it may return connection errors even before the error channel is setup. 

We should assume that if consumerGroup.Consume returns an error, it didn't start the consumer group properly (and perhaps retry?)

slinkydeveloper:

Fixes #851

### DIFF
--- a/kafka/common/pkg/kafka/consumer_factory.go
+++ b/kafka/common/pkg/kafka/consumer_factory.go
@@ -77,8 +77,10 @@ func (c kafkaConsumerGroupFactoryImpl) StartConsumerGroup(groupID string, topics
 	consumerHandler := NewConsumerHandler(logger, handler)
 
 	go func() {
-		// If any error, they are reported in error channel, so we can skip this message
-		_ = consumerGroup.Consume(context.TODO(), topics, &consumerHandler)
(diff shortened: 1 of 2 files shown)


kafka/common/pkg/kafka/consumer_factory.go
```diff
@@ -77,8 +77,10 @@ func (c kafkaConsumerGroupFactoryImpl) StartConsumerGroup(groupID string, topics
 	consumerHandler := NewConsumerHandler(logger, handler)
 
 	go func() {
-		// If any error, they are reported in error channel, so we can skip this message
-		_ = consumerGroup.Consume(context.TODO(), topics, &consumerHandler)
+		err2 := consumerGroup.Consume(context.TODO(), topics, &consumerHandler)
```
matzew: not fancy w/ `err2` - but we have done that before

PR #906, opened 2020-02-11
enable e2e tests for kafka channel

vaikas:
## Proposed Changes

  * Enable kafka e2e tests.

**Release Note**

```release-note
```


test/e2e-tests.sh
```diff
@@ -210,7 +210,9 @@ function camel_teardown() {
 
 initialize $@ --skip-istio-addon
 
-go_test_e2e -timeout=20m -parallel=12 ./test/e2e -channels=messaging.knative.dev/v1alpha1:NatssChannel || fail_test
+# DO NOT SUBMIT
```
mattmoor: Found "DO NOT SUBMIT".

mattmoor: Found "DO NOT SUBMIT".

mattmoor: Found "DO NOT SUBMIT".

mattmoor: Found "DO NOT SUBMIT".

mattmoor: Found "DO NOT SUBMIT".

test/config/100-kafka-ephemeral-triple-2.4.0.yaml
```diff
@@ -13,20 +13,13 @@ spec:
...
-      type: jbod
-      volumes:
-        - id: 0
-          type: persistent-claim
-          size: 100Gi
-          deleteClaim: false
+      type: ephemeral
   zookeeper:
-    replicas: 1
+    replicas: 3
```
matzew: For the record: this is just needed, b/c of `ephemeral` storage usage.

To quote @scholzj, from https://github.com/strimzi/strimzi-kafka-operator/pull/2487

> The ephemeral storage is just an emptyDir and can get easily lost for example when pod restarts, when you do a rolling update etc. This kinda kills Kafka, because it looses all the information stored there. With 3 Zoo nodes, they can recover from each other. So the probability of thi happening is much smaller. Changing the node count here to 1 would probably make this much worse experience for many users using it.

Still the `ephemeral` seems "good enough" for the test env.

test/config/100-strimzi-cluster-operator-0.16.2.yaml
```diff
@@ -7,12 +7,12 @@ metadata:
 spec:
   group: kafka.strimzi.io
   versions:
```
matzew: I have a script ;-) that downloads the latest ....   and not sure if @chizhg would be intersted in that kinda stuff

vaikas: Yeah, I was mainly just experimenting with why the tests were failing. When I tried to run them on my own cluster, I could get them to pass every now and then. Looks like progress is happening. I'm going to try again with parallelism since I'm seeing some wonky things with kafka channel with multiple subscriptions...

test/e2e-tests.sh
```diff
@@ -210,7 +210,8 @@ function camel_teardown() {
...
```
mattmoor: Found "DO NOT SUBMIT".

matzew:
/lgtm

:pray:

PR #912, opened 2020-02-12
strip ce_ headers before adding extensions

### LINKED ISSUE #904: Kafka channel loses broker TTL header
## Expected Behavior
Using kafka channel out of the box should work with Broker

## Actual Behavior
It does not because it drops the TTL extension header as witnessed by the logs:
{"level":"warn","ts":"2020-02-11T11:48:00.158Z","logger":"broker_filter","caller":"filter/filter_handler.go:190","msg":"No TTL seen, dropping","commit":"b6f2702","triggerRef":"kafka-broker/display-trigger","event":"Validation: valid\nContext Attributes,\n  specversion: 1.0\n  type: dev.knative.cronjob.event\n  source: /apis/v1/namespaces/kafka-broker/cronjobsources/test-cronjob-source\n  id: 64877a8e-cf41-47fe-89a9-73d6bb94c1ff\n  time: 2020-02-11T11:48:00Z\n  datacontenttype: application/json\nData,\n  {\n    \"message\": \"Hello world!\"\n  }\n"}

  1. Create a broker that uses kafka channels as per our installation documents
  2. Create a cronjob source that sends events to it and create a trigger that sends events to event display ksvc
  3. events don't show up, and the filter-dispatcher has the log message above.

## Additional Info

vaikas:
Fixes #904 

## Proposed Changes

  * When writing CE extension attributes, prefix them with ce_ prefix and when reading, strip it out.
 * Fixes couple of bugs:
  - Because of the IsAlphaNumeric function, the extensions would *NOT* be ever round tripped through Kafka
  - Because they were prefixed on the way in, they would have had wrong values in any case.

**Release Note**

```release-note
```

### DIFF
--- a/kafka/channel/pkg/dispatcher/dispatcher.go
+++ b/kafka/channel/pkg/dispatcher/dispatcher.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -352,9 +353,13 @@ func fromKafkaMessage(ctx context.Context, kafkaMessage *sarama.ConsumerMessage)
 		case "ce_dataschema":
 			event.SetDataSchema(v)
 		default:
-			// Extensions
-			if IsAlphaNumeric(h) {
-				event.SetExtension(h, v)
+			// Extensions. Strip ce_ first before calling IsAlphaNumeric
+			// because _ is not valid...
+			// TODO: Should we even allow headers to be added as extensions
+			// if they do not have the ce_ prefix.
+			strippedHeader := strings.TrimPrefix(h, "ce_")
+			if IsAlphaNumeric(strippedHeader) {
+				event.SetExtension(strippedHeader, v)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


kafka/channel/pkg/dispatcher/dispatcher.go
```diff
@@ -352,9 +353,13 @@ func fromKafkaMessage(ctx context.Context, kafkaMessage *sarama.ConsumerMessage)
 		case "ce_dataschema":
 			event.SetDataSchema(v)
 		default:
-			// Extensions
-			if IsAlphaNumeric(h) {
-				event.SetExtension(h, v)
+			// Extensions. Strip ce_ first before calling IsAlphaNumeric
+			// because _ is not valid...
+			// TODO: Should we even allow headers to be added as extensions
```
slinkydeveloper: Not sure we should add normal kafka headers in this way, KafkaSource behaves differently: https://github.com/knative/eventing-contrib/blob/master/kafka/source/pkg/adapter/adapter.go#L280 . For now I would prefer to remove it, as it creates another inconsistency

vaikas: so, if the header does not begin with ce_ do NOT add it?

slinkydeveloper: yes OR, as i do in kafka source, prepend it as ce_kafkaheader

matzew: @slinkydeveloper I have hopes, that w/ the Kafka support directly on the CloudEvent SDK, we get rid of all of this crap

PR #913, opened 2020-02-12
diff the dispatcher.Spec and update as necessary

vaikas:
If you deploy new Kafka controller and an accompanying different Dispatcher (aka, different image), the dispatcher is never updated and the old dispatcher will keep on running and not picking up the new image.

## Proposed Changes

  * Diff the spec of the Dispatcher and update if it's different from expected.

**Release Note**

```release-note
Update dispatcher deployment if Spec changes. For example, to pick up a new image.
```

### DIFF
--- a/kafka/channel/pkg/reconciler/controller/kafkachannel_test.go
+++ b/kafka/channel/pkg/reconciler/controller/kafkachannel_test.go
@@ -390,6 +390,77 @@ func TestTopicExists(t *testing.T) {
 			dispatcherNamespace:      testNS,
 			dispatcherDeploymentName: testDispatcherDeploymentName,
 			dispatcherServiceName:    testDispatcherServiceName,
+			dispatcherImage:          testDispatcherImage,
+			kafkaConfig: &KafkaConfig{
+				Brokers: []string{brokerName},
+			},
+			kafkachannelLister: listers.GetKafkaChannelLister(),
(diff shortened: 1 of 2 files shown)


kafka/channel/pkg/reconciler/controller/kafkachannel_test.go
```diff
@@ -390,6 +390,77 @@ func TestTopicExists(t *testing.T) {
 			dispatcherNamespace:      testNS,
 			dispatcherDeploymentName: testDispatcherDeploymentName,
 			dispatcherServiceName:    testDispatcherServiceName,
+			dispatcherImage:          testDispatcherImage,
+			kafkaConfig: &KafkaConfig{
+				Brokers: []string{brokerName},
+			},
+			kafkachannelLister: listers.GetKafkaChannelLister(),
+			// TODO fix
```
matzew: wanna do an issue for that ?

vaikas: lol :) Copy and paste from an existing test, so not sure what the actual TODO fix is :) 
But, sure I'll create an issue :)

vaikas: https://github.com/knative/eventing-contrib/issues/915

PR #920, opened 2020-02-12
try with parallel=3

vaikas:
Fixes #

## Proposed Changes

  * increase parallel to 6.
  *
  *

**Release Note**

```release-note
```

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -108,7 +108,8 @@ required = [
 
 [[constraint]]
   name = "knative.dev/eventing"
-  branch = "master"
+  source = "github.com/vaikas/eventing"
+  branch = "dump-events"
 
 [[constraint]]
   name="sigs.k8s.io/controller-runtime"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -108,8 +108,7 @@ required = [
 
 [[constraint]]
   name = "knative.dev/eventing"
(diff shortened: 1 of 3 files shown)


matzew:
is that an issue w/ resources of the machines ?

vaikas:
@matzew unclear. I'm concerned that these are exposing some other issues, hence I'll keep playing with it. It might be a good idea to get more thorough / stress testing in the e2e like eventing with many triggers to see if there are some concurrency issues. Or just start with more stress testing of the channels. Not sure yet, but I think we need more tests :)

nachocano:
@vaikas I'm having this same issue with the pubsub channel. See https://github.com/google/knative-gcp/pull/590
According to @chizhg this doesn't happen to natss nor imc, so we might be doing something weird in kafka and pubsub... 
I have some cycles and will be taking a closer look at pubsub. I'll sync with you in case I see something weird that may also fix kafka...

vaikas:
@nachocano Thanks! Yeah, keep me posted if you find something, I think @matzew was going to take a deep look at the Kakfa channels in general, so hopefully we can get to the bottom of this sooner rather than later :)

pierDipi:
I think I found the bug: https://github.com/knative/eventing-contrib/pull/1165

vaikas:
Thanks @pierDipi 👍

PR #929, opened 2020-02-16
only see if image has changed before redeploy

vaikas:
## Proposed Changes

  * Deployment is getting interpreted as being out of date, because the template
    we use gets a bunch of stuff defaulted into it. SO, just look at a different image (which typically is the one that changes) and redeploy then. This is causing unnecessary churn and noise in the logs and us updating the deployment unnecessarily.
  *
  *

**Release Note**

```release-note
```

### DIFF
--- a/kafka/channel/pkg/reconciler/controller/kafkachannel.go
+++ b/kafka/channel/pkg/reconciler/controller/kafkachannel.go
@@ -401,8 +401,8 @@ func (r *Reconciler) reconcileDispatcher(ctx context.Context, dispatcherNamespac
 		logging.FromContext(ctx).Error("Unable to get the dispatcher deployment", zap.Error(err))
 		kc.Status.MarkServiceUnknown("DispatcherDeploymentFailed", "Failed to get dispatcher deployment: %v", err)
 		return nil, err
-	} else if !reflect.DeepEqual(expected.Spec, d.Spec) {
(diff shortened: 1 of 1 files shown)


kafka/channel/pkg/reconciler/controller/kafkachannel.go
```diff
@@ -401,8 +401,8 @@ func (r *Reconciler) reconcileDispatcher(ctx context.Context, dispatcherNamespac
 		logging.FromContext(ctx).Error("Unable to get the dispatcher deployment", zap.Error(err))
 		kc.Status.MarkServiceUnknown("DispatcherDeploymentFailed", "Failed to get dispatcher deployment: %v", err)
 		return nil, err
-	} else if !reflect.DeepEqual(expected.Spec, d.Spec) {
-		logging.FromContext(ctx).Info("Deployment is not what we expect it to be, updating Deployment")
+	} else if !reflect.DeepEqual(expected.Spec.Template.Spec.Containers[0].Image, d.Spec.Template.Spec.Containers[0].Image) {
```
grantr: Just noticed this going through release notes. This seems like a job for DeepDerivative.

vaikas: I could've sworn I tried that and it didn't work.

grantr: That's certainly possible :D

PR #946, opened 2020-02-20
migrate to latest pkg.

n3wscott:
## Proposed Changes

  * duckv1alpha1 status was deleted. migrate off.
  * Some behavior change with MarkTrue on conditions, fixed.
  * No real code change

**Release Note**

```release-note
NONE
```


awssqs/pkg/apis/sources/v1alpha1/aws_sqs_types.go
```diff
@@ -119,12 +116,19 @@ func (s *AwsSqsSourceStatus) InitializeConditions() {
 
 // MarkSink sets the condition that the source has a sink configured.
 func (s *AwsSqsSourceStatus) MarkSink(uri string) {
-	s.SinkURI = uri
 	if len(uri) > 0 {
-		condSet.Manage(s).MarkTrue(AwsSqsSourceConditionSinkProvided)
+		if u, err := apis.ParseURL(uri); err != nil {
+			s.SinkURI = nil
+			condSet.Manage(s).MarkUnknown(AwsSqsSourceConditionSinkProvided,
+				"SinkEmpty", "Sink resolving resulted in an error. %s", err.Error())
```
grantr: This is a bug fix right? Does it change the behavior when a referenced sink has an invalid addressable uri? If so add it to release notes.

pkg/controller/sdk/status_accessor_test.go
```diff
@@ -1,114 +0,0 @@
-/*
```
grantr: Why did these tests get deleted but not the code under test? Is that code unused? Are the tests invalid?

n3wscott: the entire sdk needs to get deleted and I did not want to migrate the tests off the v1alpha1 Status

awssqs/pkg/apis/sources/v1alpha1/aws_sqs_types.go
n3wscott: it should be exactly the same.

n3wscott: and if it was important there would be a test. Tests passed.

grantr: Users don't care about tests or not, they care about what changed. We have lots of untested behavior. But if you're sure it's the same, no need for a release note.

pkg/controller/sdk/status_accessor_test.go
grantr: IMO if the code is still used, it should have tests.

awssqs/pkg/apis/sources/v1alpha1/aws_sqs_types.go
n3wscott: I am going to delete the sdk. NOW.

pkg/controller/sdk/status_accessor_test.go
n3wscott: https://github.com/knative/eventing-contrib/issues/947

PR #963, opened 2020-02-25
Get rid of the hacky patch for test images and use sed instead

chizhg:
Fixes #

## Proposed Changes
- Get rid of the hacky patch for e2e test images, and use `sed` command instead.


coryrc:
Oh, I see

/lgtm

PR #1003, opened 2020-03-07
Le 0.13

pmorie:
Brings in knative/pkg#1142 and knative/pkg#1145:

- correct behavior when leader election config doesn't exist
- fix problem with invalid prometheus metrics names with some controllers

**Release Note**

```release-note
- correct behavior when leader election config doesn't exist
- fix problem with invalid prometheus metrics names with some controller names
```


pmorie:
Wrong branch :(

Issue #997, opened 2020-03-06
Github source doesn't seem to create webhook until kicked


mattmoor:
The first time I apply github sources they don't seem to create their webhooks.

After I touch something cosmetic (e.g. change types) and re-apply they come up fine.

cc @n3wscott @grantr

mattmoor:
I believe I found the culprit here.  This still has `.eventing` in it:
https://github.com/knative/eventing-contrib/blob/bc4f495c6c296ad4d98f6d2bc7a2276173aa38da/pkg/apis/sources/register.go#L20

This means that literally every filter call in (at least) the github source will filter away all child informer events:
https://github.com/knative/eventing-contrib/blob/bc4f495c6c296ad4d98f6d2bc7a2276173aa38da/github/pkg/reconciler/controller.go#L84

mattmoor:
This is now fixed.

PR #1007, opened 2020-03-10
Fix the apiVersion used in the GitHubSource filters.

mattmoor:
- 🐛 This fixes the filter used with the OwnerReferences on all of the GitHubSource informers (and possibly more in other contrib sources).

Fixes: https://github.com/knative/eventing-contrib/issues/997

### DIFF
--- a/pkg/apis/sources/register.go
+++ b/pkg/apis/sources/register.go
@@ -17,5 +17,5 @@ limitations under the License.
 package sources
 
 const (
-	GroupName = "sources.eventing.knative.dev"
+	GroupName = "sources.knative.dev"
 )


mattmoor:
It looks like Prometheus and CouchDb have the same problem (from a superficial inspection).

lionelvillard:
is this file needed? I actually wonder do we need anything under `pkg`? @vaikas @n3wscott @grantr @matzew ?

mattmoor:
@lionelvillard I was going with a minimal fix for cherry-picking, but I agree that we should probably eliminate these subtrees.  I'm happy to follow up with that, if we all agree.  I'll start on that now to see if there are any landmines.

mattmoor:
Going to remake this against `release-0.13`

PR #1009, opened 2020-03-10
Drop pkg/apis in eventing-contrib.

mattmoor:
This is a more complete fix for #1007 (which is intended for cherry-picking), which pushes the shared functionality into individual packages.


mattmoor:
/hold

I want to discuss whether we should land #1007 and cherrypick (then land this) or whether I should just redo #1007 against release-0.13.

couchdb/source/pkg/apis/sources/v1alpha1/register.go
```diff
@@ -28,17 +28,23 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"knative.dev/eventing-contrib/couchdb/source/pkg/apis/sources"
```
lionelvillard: maybe use `"knative.dev/eventing/pkg/apis/sources"`?

couchdb/source/pkg/apis/sources/register.go
```diff
@@ -17,5 +17,5 @@ limitations under the License.
 package sources
 
 const (
-	GroupName = "sources.eventing.knative.dev"
+	GroupName = "sources.knative.dev"
```
lionelvillard: we should probably get rid of this file and use the GroupName in `knative.dev/eventing/pkg/apis/sources`

github/pkg/apis/sources/register.go
```diff
@@ -14,5 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Package sources contains sources API versions
 package sources
+
+const (
+	GroupName = "sources.knative.dev"
```
lionelvillard: same here.

mattmoor:
Updated.  I partially reverted the first commit, so `group.go` is still there for package comments, but `register.go` is gone and the `v1alpha1` imports now pick up the constant from eventing.

lionelvillard:
thanks!
/lgtm

mattmoor:
@lionelvillard For cherry-picking, I'd be more comfortable with the original (tiny) change.  Are you ok with me preparing that as a CP against release-0.13?

/unhold

mattmoor:
ugh, need to run codegen, sec...

lionelvillard:
yup that's fine with me.

mattmoor:
Alright, RFAL with `Gopkg.lock` in a separate commit 🙏

PR #1010, opened 2020-03-10
Drop use of deprecated filter API

mattmoor:


couchdb/source/pkg/reconciler/controller.go
```diff
@@ -70,12 +70,12 @@ func NewController(
 	couchdbSourceInformer.Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
 
 	deploymentInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
-		FilterFunc: controller.Filter(v1alpha1.SchemeGroupVersion.WithKind("CouchDbSource")),
+		FilterFunc: controller.FilterGroupKind(v1alpha1.Kind("CouchDbSource")),
```
dprotaso: surprised this isn't `CouchDB`

mattmoor: Same, but that ship has sailed.

mattmoor:
Rebased

PR #1015, opened 2020-03-10
Update CouchDB to use genreconciler

mattmoor:
- 🧽 This also cleans up a few peripheral pieces of code, including the use of the Lister for fetching resources (previously we fetched thru the client).


mattmoor:
/assign @lionelvillard 
/hold

I'd really like Lionel to look at this because there are no tests 😬

lionelvillard:
fair enough. It works, thanks!

/lgtm
/approve
/unhold

Issue #995, opened 2020-03-06
Github Source doesn't update Github webhook when autoTLS completes


mattmoor:
What I'd expect is that when the scheme (or other parts) of the `.status.url` in the `ksvc` changes from `http` to `https` the github webhook is updated to the new scheme.

What I actually get is a webhook that still has an `http` scheme.

cc @n3wscott @grantr

mattmoor:
Looks like this keys off of `secure: <bool>` in spec instead of the scheme of the URL produced by serving.

cc @ZhiminXiang

PR #1023, opened 2020-03-11
Support for GitHubSource + auto-TLS

mattmoor:
This makes the `secure:` bit in spec a pointer so that we can distinguish between `false` and unspecified.  When unspecified, we rely on the scheme of the underlying Knative Service.

- 🐛 Fix bug
Fixes: https://github.com/knative/eventing-contrib/issues/995


mattmoor:
cc @ZhiminXiang

PR #1065, opened 2020-03-24
squash extra commas

duglin:
Signed-off-by: Doug Davis <dug@us.ibm.com>

When there are extra commas (like at the end of the string) Split() will return empty strings for the values, and the Kafka library doesn't like those for server names or topics. So this will remove the blanks and then generate an error if the resulting list of strings is empty.


lberk:
Thanks @duglin, I'll add them this time, but if possible, it'd be great to add new tests to cover this case with the PR!

/lgtm
/approve

PR #1069, opened 2020-03-24
Add unit tests for comma split kafka adapter

lberk:
Unit tests to match #1065

cc/ @duglin

### DIFF
--- a/kafka/source/pkg/adapter/adapter_test.go
+++ b/kafka/source/pkg/adapter/adapter_test.go
@@ -667,7 +667,7 @@ func TestSplitOnCommas(t *testing.T) {
 		},
 		{
 			"middle commas check",
-			",,url-one.kafka:9092,url-two.kafka:9092,,url-three.kafka:9092,url-four.kafka:9092,,"
+			",,url-one.kafka:9092,url-two.kafka:9092,,url-three.kafka:9092,url-four.kafka:9092,,",
 			[]string{
 				"url-one.kafka:9092",
 				"url-two.kafka:9092",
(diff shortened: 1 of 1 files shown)


kafka/source/pkg/adapter/adapter_test.go
```diff
@@ -628,3 +629,48 @@ func TestAdapterStartFailure(t *testing.T) {
...
+			[]string{
+				"url-one.kafka:9092",
+				"url-two.kafka:9092",
+				"url-three.kafka:9092",
+			},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			got := SplitOnCommas(tt.bootstrapServers)
+			if diff := cmp.Diff(tt.want, got); diff != "" {
```
duglin: just wondering why you used `cmp` instead of `reflect.DeepEqual` ?

duglin:
thanks for doing this. The only other test I would add is for `,,one,two,,three,four,,`.
Because your repeated commas case (`,,`) could work due to leading or trailing duplicate `,` checking, and not necessarily for middle `,` dup checking. This one case above should get all 3 cases in one shot.

kafka/source/pkg/adapter/adapter_test.go
lberk: Path of least surprise in the codebase
```
% git grep -i cmp.diff | grep -v ^vendor | grep _test.go | wc -l
63
% git grep -i reflect.deep | grep -v ^vendor | grep _test.go | wc -l
0

lberk:
@duglin added the extra testcase.

kafka/source/pkg/adapter/adapter_test.go
duglin: ah didn't realize we were using it in other spots - thanks for the info

duglin:
LGTM

PR #1089, opened 2020-03-29
Remove finalizer to avoid finalize loop in GitHubSource FinalizeKind().

### LINKED ISSUE #1058: Github resource deletion stuck when external webhook does not exist
**Describe the bug**
The Github finalizer is not removed when the external webhook is not found.

 **Expected behavior**
The finalizer should be removed when 404 is returned when trying to delete the webhook

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

tom24d:
Fixes #1058 

## Proposed Changes

-🐛return EventTypeNormal to remove finalizer when Secret is not found instead of plain err.
-🐛return EventTypeNormal to remove finalizer when 404 is returned instead of plain err.
-🧽change the event message to describe the behavior.

### DIFF
--- a/github/pkg/reconciler/githubsource.go
+++ b/github/pkg/reconciler/githubsource.go
@@ -18,7 +18,9 @@ package github
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	//k8s.io imports
@@ -46,6 +48,8 @@ import (
 	"knative.dev/pkg/logging"
 	pkgreconciler "knative.dev/pkg/reconciler"
 	"knative.dev/pkg/resolver"
+
+	ghclient "github.com/google/go-github/github"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


github/pkg/reconciler/githubsource.go
```diff
@@ -173,8 +173,9 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, source *sourcesv1alpha1.G
 		if err != nil {
 			source.Status.MarkNoSecrets("AccessTokenNotFound", "%s", err)
```
lionelvillard: I would make sure that the error is "NotFound" and do not retry if that the case. Here an example: https://github.com/knative/eventing/blob/master/pkg/reconciler/apiserversource/apiserversource.go#L172

github/pkg/reconciler/githubsource.go
```diff
@@ -186,8 +187,9 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, source *sourcesv1alpha1.G
 		// Delete the webhook using the access token and stored webhook ID
 		err = r.deleteWebhook(ctx, args)
 		if err != nil {
```
lionelvillard: same here.

github/pkg/reconciler/githubsource.go
tom24d: Thanks! I have used apierrors.IsNotFound().

github/pkg/reconciler/githubsource.go
tom24d: I have used errors.As() for this. Please review next commit. Do you think that is beautiful code?

PR #1107, opened 2020-04-06
Add a webhook for KafkaSource that runs alongside the controller.

mattmoor:
- 🐛 Fix bug

Previously KafkaSource implemented `apis.Defaultable` and `apis.Validatable`, but didn't run a webhook component.  This turns the controller itself into the webhook, which we can tease apart if needed later.  I am going to looks for other instances of this across contrib and try to get them fixed in a similar manner.

### DIFF
--- a/kafka/source/cmd/controller/main.go
+++ b/kafka/source/cmd/controller/main.go
@@ -17,14 +17,91 @@ limitations under the License.
 package main
 
 import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	sourcesv1alpha1 "knative.dev/eventing-contrib/kafka/source/pkg/apis/sources/v1alpha1"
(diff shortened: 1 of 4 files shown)


mattmoor:
Well I was going to add these more places, but I am seeing that many of the sources in contrib are lacking implementations of `apis.Defaultable` and `apis.Validatable`.  I'm going to open issues against source owners to correct these, but I think we need a higher bar for new sources in this repo.

lberk:
outside of the labels I think this makes sense.  Perhaps we can make the `consumerGroup` default to some UUID if it's not provided (and make it optional) now that we have this

matzew:
@mattmoor @lberk not sure if or how you did test this e2e, but with HEAD or 0.14.0 relase I can not apply the source yaml, see:

https://github.com/knative/eventing-contrib/issues/1143

matzew:
I used old yaml ... :man_facepalming:

Issue #655, opened 2019-10-16
Make consistent configuration of Kafka event source and channel


aslom:
## Expected Behavior

When I have configured Kafka brokers I expect I can reuse the configuration for Kafka sources and channels. 

Ideal behavior would be to allow reuse of existing Kafka operators that make Kafka available inside kubernetes cluster. That means developer or system integrator need to do error prone copying and pasting and sometimes translation between different configuration formats.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for? Event consumer (developer), Event producer, System Integrator

## Actual Behavior

Configuration is very different and it is not consistent. Moreover some aspects can be configured only for source and not for channel (and maybe vice versa)

**Exit Criteria**
The same configuration can be re-used for Kafka event source and channel. 

Investigate and document inconsistencies with Kafka operators.

## Steps to Reproduce the Problem

  1.  Configure Kafka broker or use of Kafka operators (for example https://operatorhub.io/?keyword=kafka or https://github.com/IBM/event-streams-topic#example )
  2. Configure Kafka source -  follow https://knative.dev/development/eventing/samples/kafka/source/
  3. Configure Kafka channel - follow https://knative.dev/development/eventing/samples/kafka/channel/
  4. Compare configuration artifacts created in previous steps
 

## Additional Info

For example inconsistencies reported see https://github.com/knative/eventing-contrib/issues/323 and https://github.com/knative/eventing-contrib/issues/643

aslom:
And maybe have one controller for both Kafka channel and source? https://github.com/knative/eventing-contrib/issues/533

matzew:
the consumer part is already reused, and yes, happy to make the channel cfg like the source ones

aslom:
@matzew @lionelvillard I have written down current state of how Knative/Kubernetes Kafka configuration is done and started on proposal for improvements - please edit it and comment: https://docs.google.com/document/d/1PLYm8amLRKmKK5FAo_34ZOKwTbYU9ooMjrjwkiPc7qA/edit?usp=sharing

aslom:
Thank you for your comments @matzew - I added replies and first stab on what could be shared config map / secret

aslom:
[...]

PR #1116, opened 2020-04-08
This creates a KafkaBinding.

mattmoor:
/hold
*I don't plan to land this for 0.14, but want to start socializing this for 0.15 and beyond.*

_For (much) broader context, see [here](https://docs.google.com/document/d/1-9--tOQDL-2KN1hb5HGr9hYDasBryvAqrFsbXMsiVf0/edit)._

This PR creates a binding to pair with KafkaSource in the spirit of the above document.  This is also (very) loosely based on prior work in github.com/mattmoor/bindings (I didn't do Kafka).

Generally the idea is that a user could configure:

```yaml
apiVersion: bindings.knative.dev/v1alpha1
kind: KafkaBinding
metadata: ...
spec:
  subject:
    apiVersion: serving.knative.dev/v1
    kind: Service
    selector:
      matchLabels:
        kafka: "true"  # or something

  # This matches KafkaSource
  bootstrapServers: foo,bar
  net:
    tls:
      enable: true
      cert:
        secretKeyRef:
          name: kafka-tls-secret
          key: tls.crt
      key:
        secretKeyRef:
          name: kafka-tls-secret
          key: tls.key
      caCert:
        secretKeyRef:
          name: kafka-tls-secret
          key: ca.crt
```

Then in my Service, I would simply need the following to get an authenticated client:

```go
import "knative.dev/eventing-contrib/kafka"

// Get a client for reading.
client, err := kafka.NewConsumer(ctx)
if err != nil { ... }

// Get a client for writing.
client, err := kafka.NewConsumer(ctx)
if err != nil { ... }

```

- 🎁 Add new feature

**Release Note**

```release-note
Introduce KafkaBinding
```

**Docs**

Needs docs, but see my note about targeting 0.15.


kafka/source/config/201-clusterrole.yaml
```diff
@@ -43,6 +43,31 @@ rules:
...
+- apiGroups:
+  - bindings.knative.dev
+  resources:
+  - kafkabindings/status
+  verbs:
+  - get
+  - update
+  - patch
+
+
```
mattmoor, line 70: Remove trailing whitespace:
```suggestion

```

mattmoor:
Note that this parallels: https://github.com/knative/eventing-contrib/pull/1113

kafka/source/config/201-clusterrole.yaml
mattmoor, line 70: Remove trailing whitespace:
```suggestion

```

slinkydeveloper:
/assign

Hi @mattmoor, thanks so much for this example implementation, it really helps to grasp the idea. 

I like the idea of providing a simple configuration for golang/sarama duo, however my main concern here is "the persona" who should use this `KafkaBinding`:

* You provide env reader only for golang/sarama, but what about other libraries out there? User should rewrite an "env" parser for the library of his choice or are we supposed to provide it to him?
* Every library it's potentially configurable in different ways, abstracting over it it's not so easy: just look at librdkafka difference with sarama. The configuration we have in `KafkaSource` makes sense for sarama, but perhaps it doesn't for other libraries or it needs to be changed to reflect the other library capabilities.
* The average Kafka developer uses a JVM language (Java/Scala/Kotlin usually) and interacts with the official Kafka client. You could argue that we still inject the configuration through env variables, but the JVM Kafka client already provides a simple configuration mechanism through `application.properties` file. Users usually mounts this file from a configmap straight into the classpath. My point is that I doubt that who uses the official JVM client will be interested in such env injection.

And another thing, in the example you use a ksvc, which doesn't make sense for an application that consumes/produces kafka messages, because it can't be scaled with serving.

lionelvillard:
Related to: https://docs.google.com/document/d/1PLYm8amLRKmKK5FAo_34ZOKwTbYU9ooMjrjwkiPc7qA/edit

@aslom

aslom:
I agree with @slinkydeveloper for majority of Kafka developers they expect application.properties or similar as discussed in https://github.com/knative/eventing-contrib/issues/655 (which should be extended to cover Kafka binding)

mattmoor:
> for majority of Kafka developers they expect application.properties or similar as discussed in #655

My goal here was merely to factor the logic we have to start the conversation.  I'm absolutely not a kafka expert, so I will happily defer to your judgement on the proper idioms here.  If there is an idiomatic way of exposing this same information, then we should absolutely adopt that mechanism rather than inventing something new.

Honestly, my ideal world is one where the way we bind the information works with unmodified clients (dare to dream), but that only world in cases that have plumbed through hooks (like Google Application Default Credentials, or `DOCKER_CONFIG`, or ...).


> You provide env reader only for golang/sarama, but what about other libraries out there

That's mostly because that's what we use now.  I could see us providing alternate implementations for other Go libraries (e.g. the one needing CGO?), but I think there's value in having a pure-Go version.  The canonical example of needing multiple implementations of this interface is multiple languages.

Perhaps I could move this to a `kafka/sarama` directory?

---
I guess the question I have here is how much we want for this change vs. for follow-up?  I'd rather keep this change somewhat mechanical, so if someone with more expertise could change some of the idioms we use to be better aligned either before (internally) or after this change that'd be great.

A couple other thing that I'd be happy to tackle, but wanted to wait for some discussion:
1. Making the Source use this internally (helps reduce duplication),
2. E2E tests (we have them for source/channel, which I so so so appreciate 😍)

slinkydeveloper:
> If there is an idiomatic way of exposing this same information, then we should absolutely adopt that mechanism rather than inventing something new.

My worry is that I'm not sure such mechanism exists for kafka, unless you want to allow configuration of basic things (kafka cluster to connect to, topic to send messages). For all the other configurations, it's really client specific. This includes auth configuration, retry mechanisms, definition of things like what partitioner to use, what consumer group partition assignor to use, all of these are specific to the clients. And it's really common to configure such things when you use kafka.
Unfortunately, Kafka is not so easy as HTTP where the configuration to "bind" two services is usually as simple as the target URL.

mattmoor:
@slinkydeveloper I think in that case it's harmless to have our own schema (or adopt the "most used", for some subjective definition) and surface bindings for clients that pull that information in appropriately.

kafka/source/config/201-clusterrole.yaml
mattmoor, line 70: Remove trailing whitespace:
```suggestion

```

mattmoor:
I've updated this PR so that the event that is generated to trigger the source is done via a Job bound through the Binding in a new test image.

matzew:
@mattmoor  the release notes are wrong, mention `GithubBinding`

kafka/client.go
```diff
@@ -0,0 +1,175 @@
...
+		if err != nil {
+			return nil, err
+		}
+		cfg.Net.TLS.Config = tlsConfig
+	}
+
+	return sarama.NewClient(env.BootstrapServers, cfg)
+}
+
+func NewProducer(ctx context.Context) (sarama.Client, error) {
```
slinkydeveloper, line 82: To me these two functions looks more or less the same... what if this code more than having `NewConsumer`/`NewProducer`, has a `NewConfig` that then the KafkaSource or any other binding user can eventually customize?

test/e2e/helpers/kafka_helper.go
```diff
@@ -37,73 +42,81 @@ var (
...
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: cgName,
-					},
-				}},
-			}},
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Name:  cgName + "-producer-container",
```
slinkydeveloper, line 63: I understand that you want to test your binding here, but maybe can we have it in another test? like another e2e test binding -> kafka source.
I would love to keep the kafkacat sender because it's a rock solid kafka client and helps to check the effective integration of kafka source with outside world.

matzew: +1 on that @slinkydeveloper

matzew:
/hold

doesn't this add some complexity to the source? especially since kafka is different than http.. :thinking:

mattmoor:
> doesn't this add some complexity to the source? especially since kafka is different than http

@matzew I don't follow the context for this comment, I haven't changed the source really at all (some superficial stuff at the Go-level, but nothing visible).

kafka/client.go
mattmoor: I factored a `NewConfig` with much thinner versions of these functions as "sugar".

kafka/source/pkg/adapter/adapter_benchmark_test.go
```diff
@@ -72,18 +72,23 @@ func BenchmarkHandle(b *testing.B) {
 
 	statsReporter, _ := source.NewStatsReporter()
 
+	cfg := sarama.NewConfig()
+	client, err := sarama.NewClient([]string{"server1", "server2"}, cfg)
```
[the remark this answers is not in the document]
mattmoor: done

test/e2e/helpers/kafka_helper.go
mattmoor: I'm happy to split this into two tests.  I may look at making this part of the table test, if that's plausible.  Will circle back after I do that.

slinkydeveloper:
@mattmoor did i understood correctly that this PR makes KafkaSource dependant on KafkaBinding?

mattmoor:
@slinkydeveloper It does not, but that is a direction we could take things if we wanted.

mattmoor:
From my comment above:
> A couple other thing that I'd be happy to tackle, but wanted to wait for some discussion:
> 1. Making the Source use this internally (helps reduce duplication),
> 2. E2E tests (we have them for source/channel, which I so so so appreciate 😍)

I have done `2.` now, but not `1.`.  This change is big enough.

test/e2e/helpers/kafka_helper.go
mattmoor: Alright, this should be done.  The table test was just awkward enough that it probably wasn't worth it.

slinkydeveloper:
@mattmoor ah i see, in that case i think it's fine since it just moves some client bootstrap stuff in another module.

n3wscott:
> doesn't this add some complexity to the source? especially since kafka is different than http.. 🤔

It does not. It removes complexity.

/lgtm
/approve

test/e2e/kafka_binding_test.go
```diff
@@ -0,0 +1,124 @@
...
+			Kind:       "Job",
+			Selector: &metav1.LabelSelector{
+				MatchLabels: selector,
+			},
+		},
+	))
+
+	client.WaitForAllTestResourcesReadyOrFail()
+
+	helpers.MustPublishKafkaMessageViaBinding(client, selector, kafkaTopicName, messageKey, messageHeaders, messagePayload)
```
slinkydeveloper, line 74: :+1:

kafka/client.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	cfg.Consumer.Offsets.Initial = sarama.OffsetOldest
+	cfg.Consumer.Return.Errors = true
+
+	return sarama.NewClient(bs, cfg)
+}
+
+// NewProducer is a helper method for constructing a client for producing kafka methods.
+func NewProducer(ctx context.Context) (sarama.Client, error) {
```
slinkydeveloper, line 94: Since we are in the CloudEvents world, would it be interesting to have an additional method that returns a Cloudevents Client configured for Kafka?

slinkydeveloper: This is an example for a producer: https://github.com/cloudevents/sdk-go/blob/master/v2/cmd/samples/kafka/sender/main.go#L14

And for consumer: https://github.com/cloudevents/sdk-go/blob/master/v2/cmd/samples/kafka/receiver/main.go#L15

mattmoor: This is all about making things simpler, but perhaps as a follow-up?

slinkydeveloper: Sure!

mattmoor:
I have rebased this now that the GithubBinding has landed.

mattmoor:
CamelSource :-/

/retest

PR #1118, opened 2020-04-08
GitLabSource implements apis.Validatable and apis.Defaultable, #1108

### LINKED ISSUE #1108: GitlabSource should implement apis.Validatable and apis.Defaultable
cc @sebgoa @tzununbekov 

Then we should set up the webhooks for these source types like in this PR: https://github.com/knative/eventing-contrib/pull/1107

tzununbekov:
Fixes #1108


github/pkg/apis/sources/v1alpha1/githubsource_defaults_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
-func TestCouchDbDefaults(t *testing.T) {
```
matzew, line 26: :smile_cat:

PR #1125, opened 2020-04-09
Include empty validation callback lists

whaught:
Issue https://github.com/knative/pkg/issues/1140

## Proposed Changes

- Include empty validation callback lists in eventing-contrib
- This will allow us later to remove the variadic from /pkg and  require the parameter


whaught:
/hold cancel

Added the issue to the description

PR #1175, opened 2020-04-28
Use config from context in controller initialization.

markusthoemmes:
## Proposed Changes

Instead of passing a manually fetched config, we can now use the config that comes with the passed context to initialize controllers that need a config. Cuts down on boilerplate.

/assign @n3wscott


n3wscott:
rad.

/lgtm
/approve

PR #1173, opened 2020-04-28
[github] add permission to read namespaces to the github-webhook (#1172)

Fabian-K:
Fixes #1172
This PR grants the github-webhook service account permission to read namespaces

As this is my first contribution, I hope that I did not miss something. Looking forward to your feedback!

PR #1178, opened 2020-04-28
GithubSource webhook role fix

tzununbekov:
GithubSource webhook cannot access `namespace` resource after it was moved to separate deployment:

```
E0428 15:10:27.239494       1 reflector.go:123] runtime/asm_amd64.s:1373: Failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:knative-sources:github-webhook" cannot list resource "namespaces" in API group "" at the cluster scope
```

### DIFF
--- a/github/config/201-clusterrole.yaml
+++ b/github/config/201-clusterrole.yaml
@@ -83,19 +83,6 @@ rules:
   - configmaps
   verbs: *everything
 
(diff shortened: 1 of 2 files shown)


tzununbekov:
Oopsie, dup for #1173

PR #1192, opened 2020-05-05
Updated eventing

slinkydeveloper:
Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

Updated eventing to reflect changes from https://github.com/knative/eventing/pull/3080

PR #1193, opened 2020-05-05
Fix update-deps --upgrade

mattmoor:
This also runs it.


slinkydeveloper:
@mattmoor you need to fix imports for the changes to the `MainMessageAdapter`, like done here: https://github.com/knative/eventing-contrib/pull/1192

PR #1212, opened 2020-05-08
Release WebSocket Sources image

matzew:
## Proposed Changes

- currently we have no image build for the websocket source. This leads to people not seeing it, like in the first community meeting. Folks did build a _similar_ source, used as `ContainerSource`. With this PR we will get a visible released yaml, and the image

### DIFF
--- a/config/tools/websocket-source/websocket-source.yaml
+++ b/config/tools/websocket-source/websocket-source.yaml
@@ -1,0 +1,26 @@
+# Copyright 2020 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


config/tools/websocket-source/websocket-source.yaml
```diff
@@ -0,0 +1,25 @@
...
+apiVersion: sources.knative.dev/v1alpha2
+kind: ContainerSource
+metadata:
+  name: websocket-source
+spec:
+  image: ko://knative.dev/eventing-contrib/cmd/websocketsource
+  sink:
+    apiVersion: serving.knative.dev/v1
+    kind: Service
+    name: event-display
```
Abd4llA, line 25: Do you want to call it `event-display` or this was overlooked?

lionelvillard: missing ref?

```
sink:
  ref:
    ...
```

lionelvillard:
/lgtm
/hold

remove hold after adding ref

PR #1223, opened 2020-05-13
Fixed kafka source after update to eventing

### LINKED ISSUE #1209: Re enable kafka e2e tests
Unable to fix kafka e2e tests while updating deps. Follow up with a fix and re-enable. 

https://github.com/knative/eventing-contrib/pull/1208/files#r421898921

**Describe the bug**

 **Expected behavior**

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

slinkydeveloper:

Fixes #1209

### DIFF
--- a/kafka/source/pkg/reconciler/source/resources/receive_adapter.go
+++ b/kafka/source/pkg/reconciler/source/resources/receive_adapter.go
@@ -57,7 +57,7 @@ func MakeReceiveAdapter(args *ReceiveAdapterArgs) *v1.Deployment {
 		Name:  "KAFKA_NET_TLS_ENABLE",
 		Value: strconv.FormatBool(args.Source.Spec.Net.TLS.Enable),
 	}, {
-		Name:  "SINK_URI",
+		Name:  "K_SINK",
 		Value: args.SinkURI,
 	}, {
 		Name:  "NAME",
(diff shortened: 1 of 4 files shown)


kafka/source/pkg/reconciler/source/resources/receive_adapter.go
```diff
@@ -57,7 +57,7 @@ func MakeReceiveAdapter(args *ReceiveAdapterArgs) *v1.Deployment {
 		Name:  "KAFKA_NET_TLS_ENABLE",
 		Value: strconv.FormatBool(args.Source.Spec.Net.TLS.Enable),
 	}, {
-		Name:  "SINK_URI",
+		Name:  "K_SINK",
```
matzew, line 60: that is the SinkBinding feature ... 

Do we no longer need the new ?   or was that all changed ?

slinkydeveloper: I can't really say why this env changed, i just switched from adapter to adapter/v2 and this happened :)

lberk:
after going through and properly reviewing/testing it
/lgtm

kafka/source/pkg/reconciler/source/resources/receive_adapter.go
lberk: right -- just to elaborate on that, the `K_SINK` is a var in the `adapter.EnvConfig` and it gets resolved to the sink, so that's why it still properly outputs in the logging.

Issue #1111, opened 2020-04-06
PrometheusSource should implement apis.Validatable and apis.Defaultable


mattmoor:
cc @syedriko @matzew 

Then we should set up the webhooks for these source types like in this PR: https://github.com/knative/eventing-contrib/pull/1107

MIBc:
I want to try.

PR #1225, opened 2020-05-13
Add PrometheusSource webhook

MIBc:
Fixes: #1111

## Proposed Changes
- Add validation and default webhook for prometheus source.

### DIFF
--- a/prometheus/cmd/webhook/main.go
+++ b/prometheus/cmd/webhook/main.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


prometheus/cmd/webhook/main.go
```diff
@@ -0,0 +1,105 @@
...
+		true,
+
+		// Extra validating callbacks to be applied to resources.
+		callbacks,
+	)
+}
+
+func main() {
+	ctx := webhook.WithOptions(signals.NewContext(), webhook.Options{
+		ServiceName: "prometheus-source-webhook",
```
lberk, line 95: we use `logconfig.WebhookName()` below, could we please use it here too to stay consistent?

prometheus/config/500-webhook-config.yaml
```diff
@@ -0,0 +1,45 @@
+# Copyright 2019 The Knative Authors
```
lberk, line 1: 2019?

prometheus/cmd/webhook/main.go
MIBc: Yes

MIBc: done

prometheus/config/500-webhook-config.yaml
MIBc: done

Issue #1199, opened 2020-05-06
Port Github source to adapter/v2


slinkydeveloper:
**Problem**
Github source still uses `knative.dev/eventing/pkg/adapter` module, which uses the cloudevents/sdk-go v1. We should port it to `knative.dev/eventing/pkg/adapter/v2`, which uses the new sdk-go

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)**
Event developer/Event producer

**Exit Criteria**
Github source should not use anymore `knative.dev/eventing/pkg/adapter`

**Time Estimate (optional):**
1

MIBc:
I can work on it.

Abd4llA:
@MIBc are you working on this? Just checking since it's not assigned and we wanted to move #3103

MIBc:
Yes. I will commit it as soon as possible.

PR #1230, opened 2020-05-14
Port Github source to adapter/v2

MIBc:
Fixes: #1199

## Proposed Changes

- Using v2 adapter in github source.

### DIFF
--- a/github/pkg/adapter/adapter.go
+++ b/github/pkg/adapter/adapter.go
@@ -19,79 +19,213 @@ package adapter
 import (
 	"context"
 	"fmt"
-	"log"
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
-	cloudevents "github.com/cloudevents/sdk-go"
-
-	gh "gopkg.in/go-playground/webhooks.v5/github"
 	sourcesv1alpha1 "knative.dev/eventing-contrib/github/pkg/apis/sources/v1alpha1"
-	"knative.dev/eventing/pkg/kncloudevents"
+	"knative.dev/eventing/pkg/adapter/v2"
+	"knative.dev/pkg/logging"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"go.uber.org/zap"
+	gh "gopkg.in/go-playground/webhooks.v5/github"
 )
 
 const (
-	GHHeaderEvent    = "GitHub-Event"
-	GHHeaderDelivery = "GitHub-Delivery"
+	GHHeaderEvent    = "X-GitHub-Event"
+	GHHeaderDelivery = "X-GitHub-Delivery"
 )
 
-// Adapter converts incoming GitHub webhook events to CloudEvents
-type Adapter struct {
-	client cloudevents.Client
-	source string
+var validEvents = []gh.Event{
+	gh.CheckSuiteEvent,
+	gh.CommitCommentEvent,
+	gh.CommitCommentEvent,
+	gh.CreateEvent,
+	gh.DeleteEvent,
+	gh.DeploymentEvent,
+	gh.DeploymentStatusEvent,
+	gh.ForkEvent,
+	gh.GollumEvent,
+	gh.InstallationEvent,
+	gh.IntegrationInstallationEvent,
+	gh.IssueCommentEvent,
+	gh.IssuesEvent,
+	gh.LabelEvent,
+	gh.MemberEvent,
+	gh.MembershipEvent,
+	gh.MilestoneEvent,
+	gh.OrganizationEvent,
+	gh.OrgBlockEvent,
+	gh.PageBuildEvent,
+	gh.PingEvent,
+	gh.ProjectCardEvent,
+	gh.ProjectColumnEvent,
+	gh.ProjectEvent,
+	gh.PublicEvent,
+	gh.PullRequestEvent,
+	gh.PullRequestReviewEvent,
+	gh.PullRequestReviewCommentEvent,
+	gh.PushEvent,
+	gh.ReleaseEvent,
+	gh.RepositoryEvent,
+	gh.StatusEvent,
+	gh.TeamEvent,
+	gh.TeamAddEvent,
+	gh.WatchEvent,
 }
 
-// New creates an adapter to convert incoming GitHub webhook events to CloudEvents and
-// then sends them to the specified Sink
-func New(sinkURI, ownerRepo string) (*Adapter, error) {
-	a := new(Adapter)
-	var err error
-	a.client, err = kncloudevents.NewDefaultClient(sinkURI)
-	if err != nil {
-		return nil, err
+type envConfig struct {
+	adapter.EnvConfig
+
+	// Environment variable containing GitHub secret token
+	EnvSecret string `envconfig:"GITHUB_SECRET_TOKEN" required:"true"`
+	// Environment variable containing the HTTP port
+	EnvPort string `envconfig:"PORT" default:"8080"`
+	// Environment variable containing information about the origin of the event
+	EnvOwnerRepo string `envconfig:"GITHUB_OWNER_REPO" required:"true"`
+}
+
+// NewEnvConfig function reads env variables defined in envConfig structure and
+// returns accessor interface
+func NewEnvConfig() adapter.EnvConfigAccessor {
+	return &envConfig{}
+}
+
+// gitHubAdapter converts incoming GitHub webhook events to CloudEvents
+type gitHubAdapter struct {
+	logger      *zap.SugaredLogger
+	client      cloudevents.Client
+	secretToken string
+	port        string
+	source      string
+}
+
+// NewAdapter returns the instance of gitHubReceiveAdapter that implements adapter.Adapter interface
+func NewAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, ceClient cloudevents.Client) adapter.Adapter {
+	logger := logging.FromContext(ctx)
+	env := processed.(*envConfig)
+
+	return &gitHubAdapter{
+		logger:      logger,
+		client:      ceClient,
+		port:        env.EnvPort,
+		secretToken: env.EnvSecret,
+		source:      sourcesv1alpha1.GitHubEventSource(env.EnvOwnerRepo),
 	}
-	source := sourcesv1alpha1.GitHubEventSource(ownerRepo)
-	// Check at startup if it's not a URLRef and return an error in that case.
-	src := cloudevents.ParseURLRef(source)
+}
+
+// Start implements adapter.Adapter
+func (a *gitHubAdapter) Start(stopCh <-chan struct{}) error {
+	src := cloudevents.ParseURIRef(a.source)
 	if src == nil {
-		return nil, fmt.Errorf("invalid source for github events: %s", source)
+		return fmt.Errorf("invalid source for github events: %s", a.source)
 	}
-	a.source = source
-	return a, nil
+	done := make(chan bool, 1)
+	hook, err := gh.New(gh.Options.Secret(a.secretToken))
+	if err != nil {
+		return fmt.Errorf("cannot create gitHub hook: %v", err)
+	}
+
+	server := &http.Server{
+		Addr:    ":" + a.port,
+		Handler: a.newRouter(hook),
+	}
+
+	go gracefullShutdown(server, a.logger, stopCh, done)
+
+	a.logger.Infof("Server is ready to handle requests at %s", server.Addr)
+	if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		return fmt.Errorf("could not listen on %s: %v", server.Addr, err)
+	}
+
+	<-done
+	a.logger.Infof("Server stopped")
+	return nil
+}
+
+func (a *gitHubAdapter) newRouter(hook *gh.Webhook) *http.ServeMux {
+	router := http.NewServeMux()
+	router.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		payload, err := hook.Parse(r, validEvents...)
+		if err != nil {
+			if err == gh.ErrEventNotFound {
+				w.WriteHeader(http.StatusNotFound)
+				a.logger.Info("Event not found")
+				return
+			}
+			w.WriteHeader(http.StatusBadRequest)
+			a.logger.Errorf("Error processing request: %v", err)
+			return
+		}
+		err = a.HandleEvent(payload, r.Header)
+		if err != nil {
+			a.logger.Errorf("Event handler error: %v", err)
+			w.WriteHeader(400)
+			w.Write([]byte(err.Error()))
+			return
+		}
+		a.logger.Infof("Event processed")
+		w.WriteHeader(202)
+		w.Write([]byte("accepted"))
+	})
+	return router
+}
+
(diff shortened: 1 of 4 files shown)


github/pkg/adapter/adapter.go
```diff
@@ -85,10 +91,16 @@ func (a *Adapter) handleEvent(payload interface{}, hdr http.Header) error {
 	event.SetSource(a.source)
 	event.SetSubject(subject)
 	event.SetDataContentType(cloudevents.ApplicationJSON)
```
slinkydeveloper, line 93: You don't need this line anymore

github/pkg/adapter/adapter.go
```diff
@@ -85,10 +91,16 @@ func (a *Adapter) handleEvent(payload interface{}, hdr http.Header) error {
...
+	if err := event.SetData(cloudevents.ApplicationJSON, payload); err != nil {
+		return fmt.Errorf("failed to marshal event data: %w", err)
+	}
 
-	_, _, err := a.client.Send(context.Background(), event)
-	return err
+	result := a.client.Send(context.Background(), event)
+	if !cloudevents.IsACK(result) {
+		log.Printf("Cloud Event delivery error: %v", result)
+		return fmt.Errorf("failed to send cloudevent: %v", result)
```
antoineco, line 101: `%v` -> `%w`
so that the caller can unwrap the result if needed.

github/pkg/adapter/adapter.go
```diff
@@ -47,13 +46,20 @@ type Adapter struct {
...
 	var err error
-	a.client, err = kncloudevents.NewDefaultClient(sinkURI)
+	p, err := cloudevents.NewHTTP(cloudevents.WithTarget(sinkURI))
+	if err != nil {
+		log.Fatalf("failed to create http protocol: %s", err.Error())
+	}
+	a.client, err = cloudevents.NewClient(p, cloudevents.WithUUIDs(), cloudevents.WithTimeNow())
+	if err != nil {
+		log.Fatalf("failed to create client: %s", err.Error())
+	}
```
antoineco, line 56: We have a mix of fatal and returned errors in this function.
Since the function signature includes an error return value, we should always return the error, and let the caller fail however it wants to fail.

github/pkg/adapter/adapter.go
```diff
@@ -47,13 +46,20 @@ type Adapter struct {
 func New(sinkURI, ownerRepo string) (*Adapter, error) {
 	a := new(Adapter)
 	var err error
```
antoineco, line 48: You can remove this since you assign to a new `err` below.

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
...
+		server := httptest.NewServer(router)
+		defer server.Close()
 
-		t.Run(tc.name, tc.runner(t, *ra))
+		t.Run(tc.name, tc.runner(t, server.URL, ce))
 	}
 }
 
+// mockTransport is a simple fake HTTP transport
+type mockTransport func(req *http.Request) (*http.Response, error)
```
slinkydeveloper, line 463: Looks like mockTransport is dead code

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
...
-	}
-
-	if !wantErr && err != nil {
-		return fmt.Errorf("want no error, got %v", err)
-	}
-
-	if err != nil {
-		if diff := cmp.Diff(tc.wantErrMsg, err.Error()); diff != "" {
-			return fmt.Errorf("incorrect error (-want, +got): %v", diff)
+		if err := tc.validateAcceptedPayload(t, ceClient); err != nil {
```
slinkydeveloper, line 494: Can you add asserts here also for ce attributes like subject, id, type, source, type?

slinkydeveloper:
/hold

@MIBc can you try the source to check if it's working? I think the receive adapter implementation could break if you don't pass some/all variables like `K_SINK`, `NAME`, `NAMESPACE`, `METRICS_DOMAIN`.

Check as always the GitlabSource for this :smile:

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
...
-		headers := rv.Headers
-		for n, v := range headers {
-			delete(headers, n)
-			ln := strings.ToLower(n)
-			if _, present := unimportantHeaders[ln]; !present {
-				headers[ln] = v
-			}
-		}
+	if string(got) != string(data) {
+		return fmt.Errorf("Expected %q event to be sent, got %q", data, string(got))
```
antoineco, line 513: `data` will be printed as a byte array here unless you cast it to string as well.

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
...
+}
+
+func TestServer(t *testing.T) {
+	for _, tc := range testCases {
+		ce := adaptertest.NewTestClient()
+		adapter := newTestAdapter(t, ce)
+		hook, err := gh.New()
 		if err != nil {
-			t.Fatal(err)
+			t.Error(err)
```
antoineco, line 452: `Fatal` was correct. If this step fails, the rest of the test has no chance to succeed.

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
...
 			eventID = tc.eventID
 		}
-		hdr := http.Header{}
-		hdr.Set("X-GitHub-Event", tc.eventType)
-		hdr.Set("X-GitHub-Delivery", eventID)
-		evtErr := ra.handleEvent(tc.payload, hdr)
-
-		if err := tc.verifyErr(evtErr); err != nil {
+		req.Header.Set("X-GitHub-Event", tc.eventType)
+		req.Header.Set("X-GitHub-Delivery", eventID)
```
antoineco, line 487: "X-GitHub-Event" -> `GHHeaderEvent`
"X-GitHub-Delivery" -> `GHHeaderDelivery`

github/pkg/adapter/adapter_test.go
antoineco, line 494: Instead of returning an error here just to call `t.Error` on it, you could simply fail _inside_ the `validateAcceptedPayload` function using `t.Error` or `t.Fatal`. This is a very common pattern in tests.

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
...
-	}
-
-	h := &fakeHandler{
-		//handler: tc.sink,
-		handler: sinkAccepted, // No tests expect the sink to do anything interesting
+func (tc *testCase) validateAcceptedPayload(t *testing.T, ce *adaptertest.TestCloudEventsClient) error {
```
antoineco, line 500: This function is a test helper, so it should start with `t.Helper()` so that the lines where error occur are correctly reported in logs.

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
...
-}
+	got := ce.Sent()[0].Data()
 
-func (h *fakeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	body, err := ioutil.ReadAll(r.Body)
+	data, err := json.Marshal(tc.payload)
 	if err != nil {
-		http.Error(w, "can not read body", http.StatusBadRequest)
-		return
+		return fmt.Errorf("Could not marshal sent payload: %v", err)
```
antoineco, line 509: Like I said in a previous comment, it would be better to fail here by calling `t.Fatal` instead of returning an error.

github/pkg/adapter/adapter_test.go
```diff
@@ -407,231 +409,108 @@ var testCases = []testCase{
 	},
 }
 
-// mockTransport is a simple fake HTTP transport
-type mockTransport func(req *http.Request) (*http.Response, error)
+func newTestAdapter(t *testing.T, ce cloudevents.Client) *gitHubAdapter {
```
antoineco, line 412: The `t` variable is unused.

github/pkg/adapter/adapter_test.go
```diff
@@ -18,27 +18,29 @@ package adapter
...
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"strconv"
-	"strings"
 	"testing"
+	"time"
 
 	"knative.dev/eventing-contrib/github/pkg/apis/sources/v1alpha1"
```
antoineco, line 29: Let's group this import together with the other third-party packages. Or even better: group all `knative.dev` imports together.

github/pkg/adapter/adapter.go
```diff
@@ -19,79 +19,213 @@ package adapter
...
 	"context"
 	"fmt"
-	"log"
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
-	cloudevents "github.com/cloudevents/sdk-go"
+	sourcesv1alpha1 "knative.dev/eventing-contrib/github/pkg/apis/sources/v1alpha1"
```
antoineco, line 27: Let's group this import together with the other third-party packages. Or even better: group all `knative.dev` imports together.

MIBc:
> /hold
> 
> @MIBc can you try the source to check if it's working? I think the receive adapter implementation could break if you don't pass some/all variables like `K_SINK`, `NAME`, `NAMESPACE`, `METRICS_DOMAIN`.
> 
> Check as always the GitlabSource for this

All right. I will test it asap.

github/pkg/adapter/adapter.go
MIBc: Thank for your review. I wil think about them.  :)

github/pkg/adapter/adapter_test.go
MIBc: got it

github/pkg/adapter/adapter_test.go
MIBc: done.

github/pkg/adapter/adapter_test.go
MIBc: done.

github/pkg/adapter/adapter_test.go
MIBc: Done.

github/pkg/adapter/adapter_test.go
MIBc: Done.

github/pkg/adapter/adapter_test.go
MIBc: Done.

github/pkg/adapter/adapter_test.go
MIBc: Done.

github/pkg/adapter/adapter_test.go
MIBc: Done.

github/pkg/adapter/adapter_test.go
MIBc: Done.

github/pkg/adapter/adapter.go
MIBc: Done.

github/pkg/adapter/adapter_test.go
MIBc: The` t` is used to set fake context.

MIBc:
I have fixed the code. It can run normally. @slinkydeveloper

github/pkg/adapter/adapter_test.go
antoineco: Right, I didn't even see that :)

github/pkg/adapter/adapter_test.go
```diff
@@ -402,236 +430,115 @@ var testCases = []testCase{
...
-	gotSource := event.Source()
-	if testSource != gotSource {
-		return nil, fmt.Errorf("want source %s, got %s", testSource, gotSource)
-	}
+		defer resp.Body.Close()
 
-	gotSubject := event.Context.GetSubject()
-	if tc.wantCloudEventSubject != "" && tc.wantCloudEventSubject != gotSubject {
-		return nil, fmt.Errorf("want subject %s, got %s", tc.wantCloudEventSubject, gotSubject)
+		t.Helper()
```
antoineco, line 509: This should be inside the `validateAcceptedPayload`, on the very first line of the function.

MIBc: Done.

github/pkg/adapter/adapter.go
```diff
@@ -19,79 +19,213 @@ package adapter
...
+	}
 
-	log.Printf("Handling %s", gitHubEventType)
+	a.logger.Infof("Handling %s", gitHubEventType)
 
 	cloudEventType := sourcesv1alpha1.GitHubEventType(gitHubEventType)
-	subject := subjectFromGitHubEvent(gh.Event(gitHubEventType), payload)
+	subject := subjectFromGitHubEvent(gh.Event(gitHubEventType), payload, a.logger)
 
 	event := cloudevents.NewEvent(cloudevents.VersionV1)
```
slinkydeveloper, line 212: Can you remove the `VersionV1` in order to use always the latest version? just invoke `NewEvent` without parameters

github/pkg/adapter/adapter_test.go
```diff
@@ -402,236 +430,115 @@ var testCases = []testCase{
...
-	body   []byte
-	header http.Header
+	got := ce.Sent()[0].Data()
 
-	handler func(http.ResponseWriter, *http.Request)
-}
-
-func (h *fakeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	body, err := ioutil.ReadAll(r.Body)
+	data, err := json.Marshal(tc.payload)
```
slinkydeveloper, line 536: If it's a json always avoid byte array/string comparisons since the serialization order of the fields could be not deterministic, so the assertion below `string(got) != string(data)` could fail even if the JSONs are semantically equivalent.

Parse both tc.payload and sent event data into `interface{}` and then use `reflect.DeepEqual` or `cmp.Diff`:

```
if diff := cmp.Diff(want, got); diff != "" {
 t.Errorf("unexpected data (-want, +got) = %v", diff)
}
```

github/pkg/reconciler/source/resources/service.go
```diff
@@ -47,13 +47,24 @@ func MakeService(args *ServiceArgs) *v1.Service {
 			SecretKeyRef: args.Source.Spec.SecretToken.SecretKeyRef,
 		},
 	}, {
-		Name:  "SINK",
+		Name:  "K_SINK",
```
slinkydeveloper, line 50: :+1:

github/pkg/adapter/adapter_test.go
MIBc: Done.

github/pkg/adapter/adapter.go
MIBc: Done.

github/pkg/adapter/adapter_test.go
slinkydeveloper: You didn't parsed the json:

```
var want interface{}
json.Unmarshal(tc.payload, &want)
```

etc...

PR #1237, opened 2020-05-19
add ko.yaml to specify non root images

### LINKED ISSUE #1236: Non Root Images
**Problem**
All the eventing images are run as root. Can they be packaged as non root images like knative serving images ?

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)**
Which persona is this feature for?
* Contributors
* System Integrator
* Event producer
* Event consumer

**Exit Criteria**
Able to run a kafka source adapter in user namespace without root access

**Additional context (optional)**
I have tested running kafka source controller and adapter as non root, they seem to run fine.

itsmurugappan:
Fixes #1236 

## Proposed Changes

- Add ko.yaml
- specify default non root image


matzew:
Have you seen this from @mattmoor : 

https://github.com/knative/eventing/pull/2429

sets it for all ?

itsmurugappan:
> knative/eventing#2429

Yes. Also seen it in knative/serving. But here I wasnt sure as there was lot event sources with maintainers for each one of them. So I just included the images which I tested. If it is ok  , I can set the default to non root and if anybody has specific requirement can override for that image.

PR #1247, opened 2020-05-20
Rename deprecated filter functions to new versions.

markusthoemmes:
## Proposed Changes

As per title. The old `FilterGroup[Version]Kind` are deprecated and replaced by new ones that read a bit better.


nicolaferraro:
I've opened https://github.com/knative/eventing-contrib/issues/1250 to work on CamelSource failing tests, it seems not related to this change.

PR #1268, opened 2020-05-26
Fix data race in Gitlab unit tests

### LINKED ISSUE #1264: [flaky] gitlab/pkg/adapter.TestGracefullShutdown
### Auto-generated issue tracking flakiness of test
* **Test name**: gitlab/pkg/adapter.TestGracefullShutdown
* **Repository name**: eventing-contrib

slinkydeveloper:

Fix #1264

### DIFF
--- a/gitlab/pkg/adapter/receive_adapter.go
+++ b/gitlab/pkg/adapter/receive_adapter.go
@@ -79,13 +79,13 @@ func (ra *gitLabReceiveAdapter) Start(stopCh <-chan struct{}) error {
 		return fmt.Errorf("cannot create gitlab hook: %v", err)
 	}
 
-	var wg sync.WaitGroup
-	wg.Add(1)
-
 	server := &http.Server{
 		Addr:    ":" + ra.port,
 		Handler: ra.newRouter(hook),
 	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
 
 	go func() {
 		defer wg.Done()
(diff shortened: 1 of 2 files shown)


slinkydeveloper:
cc @antoineco

gitlab/pkg/adapter/receive_adapter.go
```diff
@@ -84,30 +85,32 @@ func (ra *gitLabReceiveAdapter) Start(stopCh <-chan struct{}) error {
 		Handler: ra.newRouter(hook),
 	}
 
-	go gracefullShutdown(server, ra.logger, stopCh, done)
+	go gracefulShutdown(server, ra.logger, stopCh, endCancel)
```
antoineco, line 88: Just a suggestion to make the code a bit easier to understand (subjectively). How about using a WaitGroup here since you just want to wait for the shutdown func to return?

```go
func (ra *gitLabReceiveAdapter) Start(stopCh <-chan struct{}) error {
    // ...

    wg := new(sync.WaitGroup)
    wg.Add(1)
    go func() {
        gracefulShutdown(server, ra.logger, stopCh)
        wg.Done()
    }()

    // ...

    wg.Wait()
}

This would avoid this context-fu.

slinkydeveloper: If i understood correctly, I think in this case we can't use a `WaitGroup` because we want to shut down the context with signals, so using `context.Background` does it

antoineco: We stop because `stopCh` gets closed, not because `endCtx` gets cancelled. Look inside `gracefulShutdown`, you are creating an entirely new Context called `ctx` for the timeout. With a WaitGroup, you can drop the `endCtx` entirely.

antoineco: There you go: https://github.com/knative/eventing/blob/master/pkg/adapter/v2/main.go#L44
This is the context you're looking for. This one gets canceled due to signals, and causes the receive on `stopCh`.

antoineco:
/lgtm
Nice catch!

Issue #1299, opened 2020-06-04
Port e2e tests to new event sender and event record images


slinkydeveloper:
**Problem**
e2e test code now uses old images event details/event logger and old event sender which uses sdk-go v1. We should port our tests to use the new test images and then prune these old images from our codebase. 

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for? Knative developer

**Exit Criteria**
e2e tests should be pruned from using old event sender and event logger/event details images

**Plan:** 

* [x] Camel source #1313
* [x] Kafka Binding #1312
* [x] Kafka Source #1311

If you're gonna take one of these tasks, please write it in the comments

**Additional context**
Look at https://github.com/knative/eventing/pull/3266 for reference on what needs to be done to port the tests

slinkydeveloper:
I'm gonna take the kafka source tests

PR #1313, opened 2020-06-11
Port CamelSource to new test images

nicolaferraro:
Part of #1299


test/e2e/camel_source_test.go
```diff
@@ -72,6 +73,12 @@ func TestCamelSource(t *testing.T) {
 									"constant": body,
 								},
 							},
+							&map[string]interface{}{
+								"set-header": &map[string]interface{}{
+									"name":     "ce-datacontenttype",
```
slinkydeveloper, line 78: if it's an http header, then this should be `content-type`

PR #1340, opened 2020-06-25
Updating deps and setting krshapedlogic=false

matzew:

Fixes #

## Proposed Changes

- Ran './hack/update-deps.sh --upgrade && ./hack/update-codegen.sh'
- Setting `krshapedlogic` explicitly to `false`, since latest PKG (see https://github.com/knative/pkg/pull/1430) has _enabled_ it by default 
-

## Next

Each source/channel maintainer can decided to actually enable it - but current behavoir matches what we had anyways.

**Release Note**

```release-note

```

**Docs**


matzew:
This is similar to @whaught did on different repos as well

PR #1366, opened 2020-07-01
KafkaChannel validation tweak

matzew:

Fixes #

## Proposed Changes

- relaxing the validation, like on the source and others in eventing

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/kafka/channel/config/300-kafka-channel.yaml
+++ b/kafka/channel/config/300-kafka-channel.yaml
@@ -23,7 +23,15 @@ metadata:
     duck.knative.dev/addressable: "true"
 spec:
   group: messaging.knative.dev
-  version: v1alpha1
+  preserveUnknownFields: false
+  validation:
+    openAPIV3Schema:
+      type: object
+        # this is a work around so we don't need to flush out the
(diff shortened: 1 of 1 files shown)


kafka/channel/config/300-kafka-channel.yaml
```diff
@@ -23,7 +23,15 @@ metadata:
     duck.knative.dev/addressable: "true"
 spec:
   group: messaging.knative.dev
-  version: v1alpha1
+  preserveUnknownFields: false
```
devguyio, line 26: from my understanding from [here](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#preserving-unknown-fields) this is like "enable pruning" , and then in line 34 you're saying "disable pruning" . Is there a reason for that?

matzew: following like all others did it, when going this route.

One example: https://github.com/knative/eventing/blob/release-0.15/config/core/resources/containersource.yaml#L37-L45

PR #1379, opened 2020-07-06
KafkaSource/KafkaBinding v1beta1 e2e

### LINKED ISSUE #1375: KafkaSource/KafkaBinding update e2e tests to use v1beta1
**Problem**
Part of https://github.com/knative/eventing-contrib/issues/1332

Update e2e tests to use new version

Example: https://github.com/knative/eventing/pull/3430

aliok:
Fixes #1375

## Proposed Changes

- E2e tests for KafkaSource and KafkaBinding v1beta1

### DIFF
--- a/test/e2e/kafka_binding_test.go
+++ b/test/e2e/kafka_binding_test.go
@@ -34,7 +34,7 @@ import (
 	contribresources "knative.dev/eventing-contrib/test/lib/resources"
 )
 
-func testKafkaBinding(t *testing.T, messageKey string, messageHeaders map[string]string, messagePayload string, expectedData string) {
+func testKafkaBinding(t *testing.T, version string, messageKey string, messageHeaders map[string]string, messagePayload string, expectedData string) {
 	client := testlib.Setup(t, true)
 
 	kafkaTopicName := uuid.New().String()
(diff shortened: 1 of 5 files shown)


aliok:
cc @nlopezgi @vaikas @lionelvillard

test/e2e/kafka_binding_test.go
```diff
@@ -47,28 +47,54 @@ func testKafkaBinding(t *testing.T, messageKey string, messageHeaders map[string
...
-			APIVersion: "batch/v1",
-			Kind:       "Job",
-			Selector: &metav1.LabelSelector{
-				MatchLabels: selector,
+	t.Logf("Creating KafkaBinding %s", version)
+	switch version {
+	case "v1alpha1":
+		contribtestlib.CreateKafkaBindingV1Alpha1OrFail(client, contribresources.KafkaBindingV1Alpha1(
+			kafkaBootstrapUrl,
+			&tracker.Reference{
```
nlopezgi, line 76: Is this object identical to the one declared in line 87? If so, consider declaring it before the switch and reusing it?

test/e2e/kafka_binding_test.go
```diff
@@ -116,8 +142,11 @@ func TestKafkaBinding(t *testing.T) {
 	}
 	for name, tc := range tests {
 		tc := tc
-		t.Run(name, func(t *testing.T) {
-			testKafkaBinding(t, tc.messageKey, tc.messageHeaders, tc.messagePayload, tc.expectedData)
+		t.Run(name+"-v1alpha1", func(t *testing.T) {
```
nlopezgi, line 145: Would it be worth the effort to have some struct with the versions to test and iterate over those here? (probably only worth if more versions need to be maintained than the current two in the future, not sure about how this is expected to evolve)

vaikas: How about we create an issue to track this and do as a follow up?

vaikas:
/lgtm
/approve

/hold
if you want to address the comments, feel free to remove the hold.

test/e2e/kafka_binding_test.go
aliok: good suggestion @nlopezgi.
things here can be extracted into a helper function and then the test functions can call the helpers.
if you don't mind, let's do that afterwards or maybe even later when we have other versions.

PR #1393, opened 2020-07-13
kafka source support for avro

### LINKED ISSUE #1319: Support AVRO decoding for kafkasource
**Problem**
it would be nice if the kafkasource CRD could take an AVRO schema registry as a parameter and decode messages coming through accordingly the same way most kafka clients do, that would save lot of time having to do it in each functions

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)**
Developers using Kafka with AVRO

**Exit Criteria**
Ensure that an AVRO encoded message is being decoded before invoking the knative function

**Time Estimate (optional):**
Golang already have multiple libraries to deal with decoding avro messages. The main logic for this feature would read the schema id and version from the binary message then fetch the schema given URL passed as parameter and decode the message before passing it to the function

itsmurugappan:
Fixes #1319

## Proposed Changes

- Add support for avro decoding of message
- add a new input for providing avro schema url 
- if the avro schema url is present it would be reached once for each schema id(comes as part of message) to get the schema
- if the url is not provided or there is some prob with the schema, decoding will not be done
- added unit tests
- e2e pending

**Release Note**

```release-note
🎁  Add support for avro decoding
```


kafka/source/pkg/adapter/message.go
```diff
@@ -161,3 +176,80 @@ func getKeyTypeMapper(keyType string) func([]byte) interface{} {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var schemaResp map[string]interface{}
+	if err := json.Unmarshal(body, &schemaResp); err != nil {
+		return nil, err
+	}
+
+
```
mattmoor, line 216: Format Go code:
```suggestion

```

mattmoor, line 216: Format Go code:
```suggestion

```

itsmurugappan:
@googlebot I consent

kafka/source/pkg/adapter/message.go
```diff
@@ -161,3 +176,79 @@ func getKeyTypeMapper(keyType string) func([]byte) interface{} {
...
+		return
+	}
+	codec, err := decoder.getCodec(int(schemaID))
+	if err != nil {
+		logger.Warn("Error getting the schema", zap.Error(err))
+		return
+	}
+	native, _, err := codec.NativeFromBinary(msg.Value[5:])
+	if err != nil {
+		logger.Warn("Error decoding message", zap.Error(err))
```
slinkydeveloper, line 245: I think this and the one below should be decoding error. This `decode` should return an error that is propagated as `ConsumerMessageToHttpRequest` error

kafka/source/pkg/adapter/message.go
```diff
@@ -161,3 +176,79 @@ func getKeyTypeMapper(keyType string) func([]byte) interface{} {
...
+	}
+
+	decoder.mutex.Lock()
+	decoder.codecMap[schemaID] = codec
+	decoder.mutex.Unlock()
+	return codec, nil
+}
+
+func (decoder *AvroDecoder) decode(msg *protocolkafka.Message, logger *zap.Logger) {
+	schemaID := binary.BigEndian.Uint32(msg.Value[1:5])
```
slinkydeveloper, line 233: you need to check the bounds of the message `len(msg.Value) >= 5`

kafka/source/pkg/adapter/message.go
```diff
@@ -161,3 +176,79 @@ func getKeyTypeMapper(keyType string) func([]byte) interface{} {
...
+	if err != nil {
+		logger.Warn("Error decoding message", zap.Error(err))
+		return
+	}
+	value, err := codec.TextualFromNative(nil, native)
+	if err != nil {
+		logger.Warn("Error decoding message", zap.Error(err))
+		return
+	}
+	msg.Value = value
```
slinkydeveloper, line 253: more than mutating the message, can you return the new bytes?

kafka/source/pkg/adapter/message.go
```diff
@@ -161,3 +176,79 @@ func getKeyTypeMapper(keyType string) func([]byte) interface{} {
...
+		return nil, err
+	}
+
+	decoder.mutex.Lock()
+	decoder.codecMap[schemaID] = codec
+	decoder.mutex.Unlock()
+	return codec, nil
+}
+
+func (decoder *AvroDecoder) decode(msg *protocolkafka.Message, logger *zap.Logger) {
```
slinkydeveloper, line 232: So the point of this decoding using avro is to convert a byte representation of avro to a textual representation (in practice JSON), correct?

kafka/source/pkg/adapter/message.go
slinkydeveloper: if len < 5, i think you should fail this decode function

kafka/source/pkg/apis/sources/v1alpha1/kafka_types.go
```diff
@@ -77,6 +77,10 @@ type KafkaSourceSpec struct {
 	// +optional
 	ConsumerGroup string `json:"consumerGroup,omitempty"`
 
+	// AvroSchemaURL is the URL where AVRO schema is located.
+	// +optional
+	AvroSchemaURL string `json:"avroSchemaURL,omitempty"`
```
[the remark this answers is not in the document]
itsmurugappan: i initially dint have in v1alpha1, but reconciler we using v1alpha1, thats y included here.

kafka/source/pkg/adapter/message.go
```diff
@@ -161,3 +176,79 @@ func getKeyTypeMapper(keyType string) func([]byte) interface{} {
...
+}
+
+func (decoder *AvroDecoder) getCodec(schemaID int) (*goavro.Codec, error) {
+	decoder.mutex.RLock()
+	if decoder.codecMap[schemaID] != nil {
+		return decoder.codecMap[schemaID], nil
+	}
+	decoder.mutex.RUnlock()
+
+	resp, err := nethttp.Get(fmt.Sprintf("%s/schemas/ids/%d", decoder.schemaURL, schemaID))
```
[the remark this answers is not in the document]
itsmurugappan: yes Confluent Schema Registry

kafka/source/pkg/adapter/message.go
itsmurugappan: yes

slinkydeveloper:
@itsmurugappan I don't think we should bound the avro functionality to the confluent schema registry specifically. I wonder if we should have a field in the spec like:

```
avro:
  registry: http://example.com
  type: confluentSchemaRegistry
```

So we might add support later for other schema registry APIs without breaking the KafkaSource CRD.

May I ask the specific use case where you find this change useful? And most important, why using kafka source + ksvc that decodes the payload using the confluent schema registry it's not enough?

lionelvillard:
Thanks for doing this! 

I'm putting a hold on this one because it needs more discussions. It seems to me this is too specific and 
as @slinkydeveloper said, providing a consistent way to "push down" (close to the source) event transformation and filtering might be the way to go. 

/hold

itsmurugappan:
@slinkydeveloper I thought this will be helpful in a couple of ways

1. To keep the ksvc's (Cloud event consumers) generic. (i.e. same consumer for different sources)
2. Will save the users from writing the same piece of logic to decode a kafka message in all the ksvc's, which includes fetching the schema id from kafka message and schema from an url.

@alex3165 please provide your thoughts as well.

kafka/source/pkg/apis/sources/v1alpha1/kafka_types.go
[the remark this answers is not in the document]
matzew: Let's get the reconciler fixed/polished first /cc @aliok

kafka/source/pkg/adapter/message.go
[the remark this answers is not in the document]
matzew: @gunnarmorling any comments on this section ?

kafka/source/pkg/apis/sources/v1alpha1/kafka_types.go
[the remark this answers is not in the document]
aliok: Ticket here: https://github.com/knative/eventing-contrib/issues/1395

kafka/source/pkg/adapter/message.go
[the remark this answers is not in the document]
matzew: Or perhaps @alesj might have some thoughts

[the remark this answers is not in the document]
gunnarmorling: Hey, so I lack a bit of insight what this change is about, but ideally it would not be based on the assumption of any specific schema registry. E.g. the Apicurio registry also supports Avro and should be usable here. While it has a compatibility mode where its API should be the same as the one exposed by Confluent Schema Registry, Apicurio also has its own API. It'd be interesting to see what it'd take to support that one here.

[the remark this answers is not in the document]
slinkydeveloper: Look at this comment below https://github.com/knative/eventing-contrib/pull/1393#issuecomment-660847433

alex3165:
I am not very familiar with Knative but I looked into introducing it in my company but having to decode the message within the cloud function seemed redundant which is why i thought having the decoding happening before invoking the function could help. By removing this barrier to entry (AVRO decoding in function) and exposing in your doc that KafkaSource CRD support AVRO decoding, it would probably bring more adoption of that CRD amongst developers using Kafka with AVRO (big proportion of the Kafka community).

If you think about Kafka clients, most of them do the AVRO decoding bit in the background and the users are just exposed to the payload of the message in a human readable format.

I agree on finding an API that doesn't bound to confluent schema registry.

aslom:
I think support for avro would be great but this PR is pretty big. @itsmurugappan any cance we can make it smaller and stage it? For example to resolve support for avro schema registries later?

itsmurugappan:
> I think support for avro would be great but this PR is pretty big. @itsmurugappan any cance we can make it smaller and stage it? For example to resolve support for avro schema registries later?

@aslom without a schema registry , we have to find a way to feed the schema , other ways that I could think of are as environment variable or config map which would make the kafka source specific to a single type of schema. Or the schema could be served by any service which serves the path /schemas/ids/<id retrieved from the kafka message> . Please let me know if there any other options.

aslom:
@itsmurugappan which one do you think is the simplest to start with? I think having kafka source limited to one schema is good start?

itsmurugappan:
sure.. let me start with one schema per kafka source and feed it as a env variable. Will make the change once the reconciler pr for v1beta1 is merged.

itsmurugappan:
@aslom have made the changes to refer the schema from env. Please review.

matzew:
@itsmurugappan sorry it's taking so long.

care to rease the PR, again?  sorry ...

kafka/source/pkg/adapter/adapter_test.go
```diff
@@ -447,3 +479,17 @@ func TestAdapter_Start(t *testing.T) { // just increase code coverage
 
 	cancel()
 }
+
+func encodedMsg(m map[string]interface{}) []byte {
```
slinkydeveloper, line 483: s/encodedMsg/mustEncodeAvroMsg

kafka/source/pkg/adapter/adapter_test.go
```diff
@@ -447,3 +479,17 @@ func TestAdapter_Start(t *testing.T) { // just increase code coverage
...
+	schemaIDBytes := make([]byte, 4)
+	binary.BigEndian.PutUint32(schemaIDBytes, uint32(1234))
+	codec, _ := goavro.NewCodec(loginEventAvroSchema)
```
slinkydeveloper, line 486: in case of error, panic using t.Fatal

kafka/source/pkg/adapter/adapter_test.go
```diff
@@ -447,3 +479,17 @@ func TestAdapter_Start(t *testing.T) { // just increase code coverage
...
+
+	binary, _ := codec.BinaryFromNative(nil, m)
```
slinkydeveloper, line 488: same as above

kafka/source/pkg/adapter/message.go
```diff
@@ -30,12 +30,17 @@ import (
...
 	"github.com/cloudevents/sdk-go/v2/extensions"
 	"github.com/cloudevents/sdk-go/v2/protocol/http"
+	"github.com/linkedin/goavro/v2"
 	"go.opencensus.io/trace"
 	"go.uber.org/zap"
 
 	sourcesv1beta1 "knative.dev/eventing-contrib/kafka/source/pkg/apis/sources/v1beta1"
 )
 
+type avroDecoder struct {
```
slinkydeveloper, line 40: Can we have this code in a separate file, or even better a separate module? like `adapter/datacodec/avro.go`?

kafka/source/pkg/apis/sources/v1beta1/kafka_types.go
```diff
@@ -77,6 +77,10 @@ type KafkaSourceSpec struct {
 	// +optional
 	ConsumerGroup string `json:"consumerGroup,omitempty"`
 
+	// AvroSchema is the AVRO schema used to decode the message.
+	// +optional
+	AvroSchema string `json:"avroSchema,omitempty"`
```
slinkydeveloper, line 82: As the comment https://github.com/knative/eventing-contrib/pull/1393#issuecomment-661044687, let's keep this on hold

itsmurugappan: this is not the schema uri, this is the schema itself.  This is as per this [comment](https://github.com/knative/eventing-contrib/pull/1393#issuecomment-664428863).

slinkydeveloper:
Ok after the chat we had in today eventing WG, IMO we should:

* change the CRD shape in order to create a field that could be brought up to the Sink duck type level. I'm thinking to something like:

``` yaml
spec:
  topics: ["topic-a"]
  transformations:
    data:
      avroDecode:
        avroSchema: "schema"
        otherAvroDecodeConfig: "blabla"
```

* We should put this on hold until we properly support https://docs.google.com/document/d/1jluAYCJi6kq6HBmR7uMRUB7Zxcf1v6qjR3-pw0gMTRU/edit?usp=sharing and release this PR as experimental feature. My worry is that we modify the "stable API" shape and then we're stuck forever with that new fields. Adding a new feature like this might require different iterations, including breaking changes (that we cannot do in a stable API), so I personally don't feel confident in adding this

matzew:
@itsmurugappan sorry to come back again.

bit this repo is no longer a thing.

Do you feel like porting your change over to the `knative-sandbox/eventing-kafka` repo? and close this issue?

itsmurugappan:
sure @matzew . I will close it.

Issue #682, opened 2019-10-28
Add Kafka Sink


aslom:
## Expected Behavior

Allow Knative sink to send events to Kafka - symmetrical to Kafka source

## Actual Behavior

Not possible with code in Knative eventing

## Steps to Reproduce the Problem

  1. Use event-display as sink to print events

## Additional Info

Based on comments @jamesward

jamesward:
Thanks!  Sorry I didn't file this.

n3wscott:
There is nothing special about a sink aside from being `addressable`. 

For example, in google/knative-gcp, it models pubsub. I created a PullSubscription backed by a deployment, and also created a Topic backed by a ksvc. A channel combines both into do what we think of the thing we want channels to do. But the components are independently useful. 

`PullSubscription` adheres to the `Source` contract, and `Topic` adheres to the `Addressable` contract. 

This is not a good first issue as it is open ended and unclear what should be produced. 

/remove-good-first-issue

aslom:
@n3wscott interesting - can I deploy only Topic component as a knative sink for gcp pubsub? Where can I find an example of it? My thinking for good first issue was that functionality is very simple could be run just with command line arguments without need to use CR.

aslom:
I wrote my own in Python as I needed it for testing - maybe I add and we refactor it as needed? I think there is real need for it ...

aslom:
First version is in PR https://github.com/knative/eventing-contrib/pull/1406

It is fully working Kafka Sink that can be deployed in k8s cluster.

Next steps are to add security, unit tests (not sure how to mock sending to Kafka topic?), create KafkaSink CR and add controller reconciliation support for it

slinkydeveloper:
Done in https://github.com/knative-sandbox/eventing-kafka-broker/pull/134

PR #1406, opened 2020-07-23
Kafka sink

aslom:
Fixes #

## Proposed Changes

Add Kafka Sink, a simple receive adapter that can be run standalone (described in README)

Ref: https://github.com/knative/eventing-contrib/issues/682

Immediately useful for users that need to get events from knative to outside Kafka broker

Current version is limited: no security support, no unit tests, no controller

Docs etc. will be addressed in future when limitations are removed and Kafka Sink was tested for some time.

### DIFF
--- a/kafka/sink/README.md
+++ b/kafka/sink/README.md
@@ -95,7 +95,8 @@ k logs -f kafka-sink-694fd8f689-smwq4
 {"level":"info","ts":"2020-07-23T16:25:09.144-0400","logger":"kafkasink","caller":"adapter/adapter.go:106","msg":"Ready to receive"}
 {"level":"info","ts":"2020-07-23T16:25:16.286-0400","logger":"kafkasink","caller":"adapter/adapter.go:109","msg":"Received message"}
 {"level":"info","ts":"2020-07-23T16:25:16.286-0400","logger":"kafkasink","caller":"adapter/adapter.go:116","msg":"Sending message to Kafka"}
-{"level":"info","ts":"2020-07-23T16:25:16.291-0400","logger":"kafkasink","caller":"adapter/adapter.go:106","msg":"Ready to receive"}```
+{"level":"info","ts":"2020-07-23T16:25:16.291-0400","logger":"kafkasink","caller":"adapter/adapter.go:106","msg":"Ready to receive"}
+```
 
 And verify that event was received by Kafka, for example
 
--- a/kafka/sink/README.md
+++ b/kafka/sink/README.md
@@ -108,10 +108,13 @@ RUn Kafka Sink receive adapter from command line:
 ```
 export KAFKA_SERVER=localhost:9092
 export KAFKA_TOPIC= test-topic
+export NAMESPACE=test
+export K_LOGGING_CONFIG=""
+export K_METRICS_CONFIG=""
 go run cmd/receive_adapter/main.go
 ```
 
-Send cloud event to Kafka Sink:
+And send an event to Kafka Sink:
 
 ```
 curl -v "http://localhost:8080" \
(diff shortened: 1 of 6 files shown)


aslom:
I followed code example provided by @slinkydeveloper and I think that current version can be useful for our users so hope we can get it merged. I listed limitations that will take longer time to implement such as creating CR and controller for Kafka Sink. 

@lionelvillard was mentioning Kafka Sink may be useful for async PoC implementation?

/assign @slinkydeveloper @lionelvillard

kafka/sink/pkg/adapter/adapter.go
```diff
@@ -0,0 +1,136 @@
...
+	client cloudevents.Client
+	//source string
+
+	port        string
+	kafkaServer string
+	topic       string
+}
+
+// NewAdapter returns the instance of gitHubReceiveAdapter that implements adapter.Adapter interface
+func NewAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, ceClient cloudevents.Client) adapter.Adapter {
```
slinkydeveloper, line 65: Why you create a source like adapter if then you don't use the provided client?

kafka/sink/cmd/receive_adapter/main.go
```diff
@@ -0,0 +1,27 @@
...
+package main
+
+import (
+	"knative.dev/eventing/pkg/adapter/v2"
+
+	kafkasinkadapter "knative.dev/eventing-contrib/kafka/sink/pkg/adapter"
+)
+
+func main() {
+	adapter.Main("kafkasink", kafkasinkadapter.NewEnvConfig, kafkasinkadapter.NewAdapter)
```
slinkydeveloper, line 26: Does it works on kube? `adapter.Main` requires a service account that can watch config maps and stuff, and from the yamls below it doesn't seem like you're using any

kafka/sink/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+The Apache Kafka Event sink enables Knative Eventing integration with Apache
+Kafka. When a CloudEvent is received over HTTP by the Knative Sink it is
+sent to the topic in Apache Kafka broker.
+
+## Deploying Kafka Sink CRDs and controller
+
+TODO: coming soon
+
+## Deploying Kafka Sink to Kubernetes
```
slinkydeveloper, line 11: I wonder, can we show an example with a sinkbinding?

kafka/sink/samples/kafka-sink-deployment.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: apps/v1
+kind: Deployment
```
matzew, line 2: would it make sense to have this thing being reflected by a CRD ?

matzew:
@aslom how about adding this to the examples? I do not really see this being "library" code.

And it would be nice if we just go w/ a deployment, to take the binding into account here

slinkydeveloper:
I was wondering, how much it takes to go from this code to a crd like:

```
type: KafkaSink
...
spec:
  topics:
    - aaa
    - bbbb
```

and then this `KafkaSink` implements the addressable duck type?

aslom:
Thank you for your comments @matzew @slinkydeveloper 

About Kafka Sink CR - yes that is the goal. However I would like to get there is smaller incremental steps getting to full functionality in small-medium size PRs instead of large XXL (that is hard for me to write and others to test).

So this is first step - MVP - Kafka sink functionality for receive adapter. Do you think this approach could work?

I started looking into Kafka Sink CR but it may require more discussion - for example could Kafka Source and Kafka Sink share the same controller?

kafka/sink/README.md
aslom: Do you mean example that shows how to use kafkasink instead of event-display? https://knative.dev/docs/eventing/samples/sinkbinding/

kafka/sink/cmd/receive_adapter/main.go
aslom: It worked in k8s v1.18 - see example output in README.md

kafka/sink/pkg/adapter/adapter.go
aslom: That is to structure it as receive adapter for future PR that add controller support and KafkaSink CR

kafka/sink/samples/kafka-sink-deployment.yaml
aslom: Yes in future

aslom:
@slinkydeveloper do you have any links (or ideas) how i could do unit test for kafka protocol (and to get code coverage pull-knative-eventing-contrib-go-coverage to pass)? Is there some way I can run unti test without requiring Kafka broker - intercept messages?

slinkydeveloper:
@aslom I think you should mock the sync producer (here you create a sender https://github.com/knative/eventing-contrib/pull/1406/files#diff-985279ee1eab4df1be848de7a44a119aR94 but you can create the same sender directly from the built sync producer). So you start the adapter, send events to it and then assert that on the other side of the producer you received something

pierDipi:
FYI, with some changes (that I will do anyway) to the `receiver` component from this project https://github.com/knative-sandbox/eventing-kafka-broker we get a Kafka Sink data plane. It already has tests (unit / integration / e2e). 
We can use it in both a multi-tenant or single-tenant setup.

aslom:
@pierDipi could you point me to kafka sink code and how you run unit tests (and do kafka sender mock)? I could not find it https://github.com/knative-sandbox/eventing-kafka-broker/search?p=2&q=kafka+sink&unscoped_q=kafka+sink

pierDipi:
It's written in Java: https://github.com/knative-sandbox/eventing-kafka-broker/tree/master/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver
Basically, the idea is to mount a config map with all configurations where each of them binds an HTTP request path (Addressable) to a Kafka topic. 
Is this the goal of a `KafkaSink`?

For testing, we use `MockProducer` for unit tests, we create an embedded Kafka cluster so that we can debug things in an IDE out of k8s for integration tests, and for e2e tests the same as in this repo.

pierDipi:
All you need to do for implementing a KafkaSink is to write a control plane that updates ConfigMap(s) depending on single-tenant vs multi-tenant setup

pierDipi:
Let me know if you find it interesting so that I can prioritize the changes I mentioned before.

aslom:
@pierDipi the goal of Kafka Sink is to be very simple to use Sink that send events ot Kafka topic: https://github.com/knative/eventing-contrib/issues/682  

I would like to make it very easy to use from CLI - the same way that event display can be deployed as a sink to show events. That is why it is just one file that listen over HTTP for CloudEvents, send them to Kafka topic. 

I would like to keep using golang to have it symmetrical to Kafka Source. 

I think mapping config maps looks  more complex than the CLI use case I have described in the github issue? However, I do not know the details how alternative kafka broker is implemented so hard to compare for me.

pierDipi:
> I would like to make it very easy to use from CLI - the same way that event display can be deployed as a sink to show events. That is why it is just one file that listen over HTTP for CloudEvents, send them to Kafka topic

> I think mapping config maps looks more complex than the CLI use case I have described in the github issue?

I don't see many differences from a usability perspective between putting configurations in a `ConfigMap` vs environment variables.

> I would like to keep using golang to have it symmetrical to Kafka Source.

So, in this case, re-using that component is not an option.

slinkydeveloper:
> I would like to keep using golang to have it symmetrical to Kafka Source.

Is there any reason for that?

aslom:
@slinkydeveloper I could not figure how to get to HTTP server started by SDK to send events for unit testing - i need to know HTTP server URL to send cloud events to it.

The closes I could find in SDK was https://github.com/cloudevents/sdk-go/blob/master/v2/client/http_receiver_test.go#L15

aslom:
@slinkydeveloper I am not able to access syncProducer in  kafka_sarama.Sender as the field is private
in protocol/kafka_sarama/v2/sender.go

```go
type Sender struct {
	topic        string
	syncProducer sarama.SyncProducer
}
```

so that code will not work even though I have `syncProducerMock`

```go
	sender := &cloudeventskafka.Sender{topic: topic, syncProducer: syncProducerMock}
```

I created PR to make possible to run

```go
	sender, err := cloudeventskafka.NewSenderFromSyncProducer(topic, syncProducerMock)
```

PR: https://github.com/cloudevents/sdk-go/pull/564

Any other/better way to run unit test then?

matzew:
@aslom I'd suggest we close this, in fav of: https://github.com/knative-sandbox/eventing-kafka-broker/pull/134

matzew:
@aslom  here is a quick sample on the code:
https://gist.github.com/matzew/e2c2fcd2696a346f25b8bc9e64bfd0fa

Let's close this?

aslom:
@matzew if I understand correctly the Kafka Sink form https://github.com/knative-sandbox/eventing-kafka-broker/pull/134 requires new broker and can not run independently? Maybe there can be more than one implementation of Kafka Sink   depending how users what to use it? In particular Kafka Sink that runs in user namespace as ksvc? @pierDipi

slinkydeveloper:
> requires new broker and can not run independently?

Not really, the data plane is abstract from the broker concept and this sink is "multi tenant"

matzew:
Just read this text:  https://gist.github.com/matzew/e2c2fcd2696a346f25b8bc9e64bfd0fa 

I do not even have the broker deployed on my cluster :smile: 

/close

Closing this now, since we have a decent type, instead of more example work, and we can improve things over there

Issue #1396, opened 2020-07-14
Make KafkaChannel v1beta1 the stored version


aliok:
Part of https://github.com/knative/eventing-contrib/issues/1350

Bump the storage version

Also think about if there's a need for a migration tool, similar to https://github.com/knative/eventing/blob/master/config/pre-install/v0.16.0/storage-version-migration.yaml

See https://github.com/knative/eventing-contrib/issues/1396 for example PRs

aliok:
/assign

This is to be done for 0.18

aliok:
/close

Moved to knative-sandbox/eventing-kafka: https://github.com/knative-sandbox/eventing-kafka/issues/70

Issue #1394, opened 2020-07-14
Make KafkaSource and KafkaBinding v1beta1 resources the stored version


aliok:
Part of https://github.com/knative/eventing-contrib/issues/1332

Bump the storage version

Also think about if there's a need for a migration tool, similar to https://github.com/knative/eventing/blob/master/config/pre-install/v0.16.0/storage-version-migration.yaml

Relevant:
- Changing storage versions: https://github.com/knative/eventing/pull/3169
- 0.16 migration job: https://github.com/knative/eventing/pull/3262
- 0.18 migration job: https://github.com/knative/eventing/pull/3895
- Docs for migration: https://github.com/knative/docs/issues/2543
- Release process updates: https://github.com/knative/eventing/pull/3283

aliok:
/assign

This is to be done for 0.18

aliok:
IMO, this depends on https://github.com/knative/eventing-contrib/issues/1482

We need to make some tooling changes (releasing pre-install.yaml, optionally implementing upgrade tests, etc.) for Kafka bits to change the storage version to v1beta1 from v1alpha1.

I would rather do these tooling changes in a clean repo to not affect other stuff. I see there's https://github.com/knative-sandbox/eventing-kafka but not sure what the directory structure there will look like.

cc @matzew @travis-minke-sap

Can we already move the KafkaChannel and KafkaSource there?

aliok:
/close

Created tickets in new repo, `knative-sandbox/eventing-kafka` for the remaining tasks: https://github.com/knative-sandbox/eventing-kafka/issues/65

PR #1504, opened 2020-08-25
WIP: change storage version to v1beta1

matzew:

Fixes #

## Proposed Changes

- setting storage version to `v1beta1`

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/kafka/channel/pkg/apis/messaging/v1alpha1/kafka_channel_defaults.go
+++ b/kafka/channel/pkg/apis/messaging/v1alpha1/kafka_channel_defaults.go
@@ -34,7 +34,7 @@ func (c *KafkaChannel) SetDefaults(ctx context.Context) {
 		c.Annotations = make(map[string]string)
 	}
 	if _, ok := c.Annotations[messaging.SubscribableDuckVersionAnnotation]; !ok {
-		c.Annotations[messaging.SubscribableDuckVersionAnnotation] = "v1alpha1"
+		c.Annotations[messaging.SubscribableDuckVersionAnnotation] = "v1beta1"
 	}
 
 	c.Spec.SetDefaults(ctx)
--- a/kafka/channel/pkg/apis/messaging/v1beta1/kafka_channel_defaults.go
+++ b/kafka/channel/pkg/apis/messaging/v1beta1/kafka_channel_defaults.go
@@ -34,7 +34,7 @@ func (c *KafkaChannel) SetDefaults(ctx context.Context) {
 		c.Annotations = make(map[string]string)
 	}
 	if _, ok := c.Annotations[messaging.SubscribableDuckVersionAnnotation]; !ok {
-		c.Annotations[messaging.SubscribableDuckVersionAnnotation] = "v1"
+		c.Annotations[messaging.SubscribableDuckVersionAnnotation] = "v1beta1"
 	}
 
 	c.Spec.SetDefaults(ctx)
--- a/kafka/channel/config/300-kafka-channel.yaml
+++ b/kafka/channel/config/300-kafka-channel.yaml
@@ -62,10 +62,10 @@ spec:
   versions:
   - name: v1alpha1
     served: true
-    storage: true
+    storage: false
   - name: v1beta1
     served: true
-    storage: false
+    storage: true
   conversion:
     strategy: Webhook
     conversionReviewVersions: ["v1beta1", "v1alpha1"]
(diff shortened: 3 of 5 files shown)


kafka/channel/pkg/apis/messaging/v1beta1/kafka_channel_defaults.go
```diff
@@ -34,7 +34,7 @@ func (c *KafkaChannel) SetDefaults(ctx context.Context) {
 		c.Annotations = make(map[string]string)
 	}
 	if _, ok := c.Annotations[messaging.SubscribableDuckVersionAnnotation]; !ok {
-		c.Annotations[messaging.SubscribableDuckVersionAnnotation] = "v1"
```
matzew, line 37: @aliok why was this at `v1`? :thinking: 

I also checked here: https://github.com/knative/eventing/pull/3169/files  and there (matching to IMC) it was at beta1...

aliok: This doesn't mean that this is the KafkaChannel version.

This is the duck type version KafkaChannel supports. So, it should be v1.

kafka/channel/pkg/apis/messaging/v1alpha1/kafka_channel_defaults.go
```diff
@@ -34,7 +34,7 @@ func (c *KafkaChannel) SetDefaults(ctx context.Context) {
 		c.Annotations = make(map[string]string)
 	}
 	if _, ok := c.Annotations[messaging.SubscribableDuckVersionAnnotation]; !ok {
-		c.Annotations[messaging.SubscribableDuckVersionAnnotation] = "v1alpha1"
+		c.Annotations[messaging.SubscribableDuckVersionAnnotation] = "v1beta1"
```
aliok, line 37: I am not so sure about this change here.
v1alpha1 Channelable duck type support is removed from eventing.

I think we should get rid of the the v1alpha1 channel resource completely, providing a migration script.

cc @vaikas for confirmation.

kafka/channel/pkg/apis/messaging/v1beta1/kafka_channel_defaults.go
aliok: > I also checked here: https://github.com/knative/eventing/pull/3169/files and there (matching to IMC) it was at beta1...

It should've been v1 too. But, there's no difference between v1beta1 and v1 Channelable/Subscribable duck type, so, v1beta1 should also work. ...until v1beta1 support is deleted.

aliok:
@matzew https://github.com/knative/eventing-contrib/issues/1396 is the ticket for this PR.
feel free to assign it to yourself.

aliok:
@matzew there's also the source and we need to do the same thing there too.

Let me create the setup for migration and then we can do the same thing for channel as well.

aliok:
More specifically:
- Example PR for storage version update and migration job: https://github.com/knative/eventing/pull/3923
- Example PR for tooling for the storage version update: https://github.com/knative/eventing/pull/3283

kafka/channel/pkg/apis/messaging/v1alpha1/kafka_channel_defaults.go
vaikas: Yes, I'd like us to get rid of the v1alpha1 shapes asap.

kafka/channel/pkg/apis/messaging/v1beta1/kafka_channel_defaults.go
vaikas: Yeah, so we have to default the annotation to the version that we are currently storing. Why? Because for different serving versions we run the conversion hook, and we can adjust the annotation to match if we fiddle with the fields. For things that we are storing, however, we can't do any modifications because the conversion hook does not run. As @aliok says, since they are the same we should be safe here, but just wanted to give the reasoning for why we set the annotations differently and what the "correct" way to do that is. Hope that helps @matzew @aliok

aliok:
About required tooling changes: https://github.com/knative/eventing-contrib/issues/1394#issuecomment-681882435

aliok:
/close

Created tickets in new repo, `knative-sandbox/eventing-kafka` for the remaining tasks: https://github.com/knative-sandbox/eventing-kafka/issues/65

PR #1675, opened 2020-12-04
Updating the file that generates the build

matzew:

Fixes #

## Proposed Changes

- config for the tool was using old syntax...

/assign @pierDipi


pierDipi (approved):
/lgtm

PR #1678, opened 2021-02-04
remove old cruft

vaikas:
## Proposed Changes

- Remove left over deps that are no longer necessary. These were missed when things were moved out to knative-sandbox repos.
-
-

**Release Note**

```release-note
:broom: remove bunch of unnecessary deps that are no longer necessary.
```

**Docs**


matzew (approved):
/lgtm
/approve

